PR #2, opened 2020-12-21
Remove macOS 10.15 platform restriction

stephencelis:
Noticed this when hunting down another thing.


mbrandonw (approved):
Nice, good catch.

PR #5, opened 2020-12-21
Add CI, Autoformat, and Delegate Floating Point Parsing

stephencelis:


Makefile
```diff
@@ -1,9 +1,12 @@
-PLATFORM_IOS = iOS Simulator,name=iPhone 11 Pro Max
+PLATFORM_IOS = iOS Simulator,name=iPhone 11 Pro
 PLATFORM_MACOS = macOS
 PLATFORM_TVOS = tvOS Simulator,name=Apple TV 4K (at 1080p)
 
 default: test
 
+benchmarks:
+	swift run swift-parsing-benchmark --allow-debug-build
```
mbrandonw, line 8:
```suggestion
	swift run -c release swift-parsing-benchmark
```

Sources/swift-parsing-benchmark/main.swift
```diff
@@ -9,15 +9,15 @@ Benchmark.main(
 //    boolSuite,
 //    colorSuite,
 //    csvSuite,
-    dateSuite,
+//    dateSuite,
 //    httpSuite,
 //    jsonSuite,
-//    numericsSuite,
```
mbrandonw, line 15: we should bring back on the benchmarks now

Tests/ParsingTests/DoubleTests.swift
```diff
@@ -50,8 +50,8 @@ final class DoubleTests: XCTestCase {
     XCTAssertEqual(" Hello", String(input))
 
     input = "1234567890123456789012345678901234567890 Hello"[...].utf8
-    XCTAssertEqual(.infinity, parser.parse(&input))
-    XCTAssertEqual(" Hello", String(input))
+    XCTAssertEqual(nil, parser.parse(&input))
+    XCTAssertEqual("1234567890123456789012345678901234567890 Hello", String(input))
```
mbrandonw, line 54: this is an expected change right?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
+name: CI
+
+on:
+  push
+  # :
+  #   branches:
+  #     - main
+  # pull_request:
+  #   branches:
+  #     - '*'
```
mbrandonw: we should uncomment these too

Tests/ParsingTests/DoubleTests.swift
stephencelis: Yeah. Overflowing to infinity is probably not what we want anyway.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - '*'
+
+jobs:
+  tests:
+    runs-on: macOS-latest
+    strategy:
+      matrix:
+        xcode:
+          - 12.3
```
mbrandonw, line 17: do we want some Xcode 11s too?

README.md
```diff
@@ -334,9 +333,9 @@ Numerics.Comma separated: Int.parser           5390299.000 ns ±   5.52 %
 Numerics.Comma separated: Scanner.scanInt     86885256.500 ns ±   5.40 %         16
 Numerics.Comma separated: String.split       120884645.000 ns ±   2.18 %         11
 Numerics.Double.init                                62.000 ns ± 615.58 %    1000000
-Numerics.Double.parser                              60.000 ns ± 640.87 %    1000000
+Numerics.Double.parser                              92.000 ns ± 249.73 %    1000000
```
stephencelis, line 336: We were using `strtod_l` and family before, which is what Swift uses under the hood and is just as fast as `Double.init`, but CI was failing to parse due to not being able to get a handle on contiguous memory. May be able to recover this again in the future, but let's be safe for now.

mbrandonw (approved):

PR #11, opened 2020-12-23
Update README.md

maximkrouk:
I've just found a `Newline` parser, so I think that's the right place to use it.

Also this PR is kind of a little feature request for `PrefixUpTo(Newline())` parser ❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ input // => "\n2,Blob Jr.,false\n3,Blob Sr.,true"
 To parse multiple users from the input we can use the `Many` parser:
 
 ```swift
-let users = Many(user, separator: StartsWith("\n"))
+let users = Many(user, separator: Newline())
 
 users.parse(&input) // => [User(id: 1, name: "Blob", isAdmin: true), ...]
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks for the PR!

I think for the purpose of the README we wanna keep things simple and as analogous to the "unrolled" version that uses `split(separator: "\n")` as possible (`Newline` introduces a lil extra work under the hood since it matches `\r\n` as well). We'll discuss, though! Maybe an opportunity to at least mention it!

maximkrouk:
Fine, also an upd on my feature request: Probably PrefixUpTo("\n") does just the same as PrefixUpTo(Newline()) would do, because both "\n" and "\r\n" end on "\n", so it's probably not as needed as I thought 🙂

stephencelis:
Hey @maximkrouk! Thanks again for the PR. I think we're going to close it for now given that we wanted to have a very simple example for the README, though your change is definitely the more _correct_ way of handling new lines. We'll keep this mind in the future when we work on any more documentation.

PR #12, opened 2020-12-24
String literal

stephencelis:
I had a tab open with [nom's example](https://github.com/Geal/nom/blob/b52d4602764a1fdee5d2602e0e0a6759b218b8e8/examples/string.rs) and decided to tackle it so I could close it. Reads pretty nicely, but also caught a bug in a few of our `Prefix` initializers, which weren't passing the predicate along 😓


Sources/Parsing/Parsers/Prefix.swift
```diff
@@ -44,7 +44,7 @@ where
   public init(
     minLength: Int = 0,
     maxLength: Int? = nil,
-    while predicate: ((Input.Element) -> Bool)? = nil
+    while predicate: @escaping (Input.Element) -> Bool
```
stephencelis, line 47: This is the one where `where` is required. Figured `Prefix()` doesn't make a lot of sense on its own.

mbrandonw (approved):

PR #13, opened 2020-12-24
Simplify parse method

kaandedeoglu:


stephencelis (approved):
Thanks for the PR! This is definitely a bit clearer given the double optional. Proper error handling should hopefully make this even clearer in the future!

PR #14, opened 2021-01-04
Add support for parsing enum raw string values

lukeredpath:
# Description

This PR adds support for parsing raw enum values off the beginning of `Input` and returning the enum value in the following ways:

* A `Parsers.enumRawValue(in:)` function that returns a parser.
* An opt-in high-level API as extensions on a custom `CaseParsable` protocol that enums can conform to.

## Use case

A common task might be to parse some kind of input into a local domain model which may frequently include some kind of enumerated type. For example, given the following string:

```swift
"thing:type_a:type_b"
```

And the following types:

```swift
struct Thing {
  var type_a: TypeA
  var type_b: TypeB
}

enum TypeA: String, CaseIterable {
  case one
}

enum TypeB: String, CaseIterable {
  case two
}
```

Using the new enum case parser you could construct a `Parser<Substring, Thing>` as follows:

```swift
let thingParser = StartsWith("thing:")
    .take(Parsers.enumRawValue(in: TypeA.allCases))
    .skip(StartsWith(":"))
    .take(Parsers.enumRawValue(in: TypeB.allCases))
    .map(Thing.init(type_a:type_b:))
```

Or, by opting the enums into the `CaseParsable` protocol the above could be rewritten as:

```swift
let thingParser = StartsWith("thing:")
    .take(TypeA.firstMatchingCase)
    .skip(StartsWith(":"))
    .take(TypeB.firstMatchingCase)
    .map(Thing.init(type_a:type_b:))
```

## Discussion

I originally implemented this in a Swift playground purely as an extension on the `RawRepresentable` protocol however I feel like libraries should avoid adding behaviour to built-in types by default and this should be opt-in, which is why the main parsing logic is implemented as a static function on `Parsers` and the high-level API implemented as default implementations on a custom protocol. Conforming to the `CaseParsable` protocol feels like a fairly unobtrusive way to gain extra functionality and is inspired by the similarly named `CaseIterable` protocol.

This PR only adds support for parsing string raw values but could be extended to support other raw value types such as `Int` although I feel like `String` would be the common use-case for this.

Main points for discussion are:

* Is implementing the base parser as a static function the best way of going about this? Because it's entirely constructed by composing existing parsers it felt overkill to create an entirely new `Parser` type. Is there a better approach?
* Naming - naming is hard, do the names of both the static function and `CaseParsable` functions feel right? I like `firstMatchingCase` although I'm not sure if it completely conveys the behaviour (that it works on the prefix of the input). I feel like `enumRawValue` could be named better.

### DIFF
--- a/Sources/Parsing/Parsers/Enum.swift
+++ b/Sources/Parsing/Parsers/Enum.swift
@@ -1,0 +1,55 @@
+extension Parsers {
+  /// Parses out the first matching raw value in a list of possible values from the start of the current input.
+  @inlinable
+  public static func enumRawValue<EnumType: RawRepresentable>(
+    in possibleValues: [EnumType]
+  ) -> AnyParser<Substring, EnumType> where EnumType.RawValue == String {
+    OneOfMany(
+      possibleValues.map { value in
+        StartsWith(value.rawValue[...]).map { value }
+      }
+    )
+    .eraseToAnyParser()
+  }
+}
+
+/// Provides an opt-in convenience high-level API for parsing enum values.
+public protocol CaseParsable: RawRepresentable {}
+
+extension CaseParsable where Self.RawValue == String {
+  /// Parses out the first matching raw value from a list of possible values, or returns nil.
+  ///
+  /// Example:
+  ///
+  ///     enum SomeValues: String, CaseParseable {
+  ///         case one
+  ///         case two
+  ///     }
+  ///
+  ///     SomeValues.firstMatchingValue(in: [.one]).parse("one").output // SomeValues.one
+  ///     SomeValues.firstMatchingValue(in: [.two]).parse("one").output // nil
+  ///
+  public static func firstMatchingCase(
+    in possibleValues: [Self]
+  ) -> AnyParser<Substring, Self> {
+    Parsers.enumRawValue(in: possibleValues)
+  }
+}
+
+extension CaseParsable where Self: CaseIterable, Self.RawValue == String {
+  /// Parses out the first matching raw value from all possible cases, or returns nil.
+  ///
+  /// Example:
+  ///
+  ///     enum SomeValues: String, CaseParseable, CaseIterable {
+  ///         case one
+  ///         case two
+  ///     }
+  ///
+  ///     SomeValues.firstMatchingValue.parse("one").output // SomeValues.one
+  ///     SomeValues.firstMatchingValue.parse("six").output // nil
+  ///
+  public static var firstMatchingCase: AnyParser<Substring, Self> {
+    Parsers.enumRawValue(in: allCases as? [Self] ?? [])
+  }
+}
--- a/Tests/ParsingTests/EnumTests.swift
+++ b/Tests/ParsingTests/EnumTests.swift
@@ -1,0 +1,33 @@
+import Parsing
+import XCTest
+
+final class EnumTests: XCTestCase {
+  enum Example: String, CaseIterable, CaseParsable {
+    case one
+    case two
+  }
+
+  func testSuccess() {
+    var input = "one"[...]
+    XCTAssertEqual(.one, Parsers.enumRawValue(in: Example.allCases).parse(&input))
+    XCTAssertEqual("", input)
+  }
+
+  func testFailure() {
+    var input = "one"[...]
+    XCTAssertNil(Parsers.enumRawValue(in: [Example.two]).parse(&input))
+    XCTAssertEqual("one", input)
+  }
+
+  func testCaseParsableWithSpecificCases() {
+    var input = "one"[...]
+    XCTAssertEqual(.one, Example.firstMatchingCase(in: [Example.one]).parse(&input))
+    XCTAssertEqual("", input)
+  }
+
+  func testCaseParsableWithAllCases() {
+    var input = "one"[...]
+    XCTAssertEqual(.one, Example.firstMatchingCase.parse(&input))
+    XCTAssertEqual("", input)
+  }
+}


stephencelis:
Hey @lukeredpath!

Thanks for bringing this up on Twitter and taking the time to PR! We think having support for `RawRepresentable`, `CaseIterable` parsing would be a great addition to the project! Just a few small design things to eke out.

> I originally implemented this in a Swift playground purely as an extension on the `RawRepresentable` protocol however I feel like libraries should avoid adding behaviour to built-in types by default and this should be opt-in, which is why the main parsing logic is implemented as a static function on Parsers and the high-level API implemented as default implementations on a custom protocol. Conforming to the `CaseParsable` protocol feels like a fairly unobtrusive way to gain extra functionality and is inspired by the similarly named `CaseIterable` protocol.

We think it's OK to extend built-in types so long as the APIs work with types we own, and in this case we own the parsing types that are vended from these endpoints. In fact we do this already with `Int.parser()`, `Double.parser()`, `Bool.parser()`, and `Optional.parser()`. Because of this I think it's safe to avoid introducing the additional, opt-in protocol and instead we can lean directly on `RawRepresentable`.

> Is implementing the base parser as a static function the best way of going about this? Because it's entirely constructed by composing existing parsers it felt overkill to create an entirely new `Parser` type. Is there a better approach?

Many of the parsers that ship with the library could be composed out of simpler parsers but are written as a single unit for additional performance. While micro- and macrotuning performance isn't usually what an everyday user of the library needs to be concerned with, we think it's a good idea to optimize the parsers that come with the library, so defining this parser as its own type with its own from-scratch `parse` method is probably the way to go. Want to try defining one, perhaps named `Parsers.RawRepresentableParser` to match `Parsers.IntParser`, etc.?

> Naming - naming is hard, do the names of both the static function and CaseParsable functions feel right? I like firstMatchingCase although I'm not sure if it completely conveys the behaviour (that it works on the prefix of the input). I feel like enumRawValue could be named better.

Naming is the hardest part for sure :smile: I think if we follow the current library conventions of `{TypeName}.parser`, we could support all of the following:

- `(RawRepresentable & CaseIterable).parser()`, _e.g._ `MyEnum.parser()`
- `RawRepresentable.parser(in: RawValue...)`, _e.g._ `MyEnum.parser(in: .case1, .case2)`
- `RawRepresentable.parser(in: [RawValue])`, _e.g._ `MyEnum.parser(in: [.case1, .case2])`

> Should this be further optimised to use UTF8 strings instead of Substring?

I think so, yup. If you look at `IntParser` we default to [generic collections of UTF-8 code units](https://github.com/pointfreeco/swift-parsing/blob/346fbdcb793c4baa845ca44664ed20f719a7d83b/Sources/Parsing/Parsers/Int.swift#L56-L58), and then have [helpers](https://github.com/pointfreeco/swift-parsing/blob/346fbdcb793c4baa845ca44664ed20f719a7d83b/Sources/Parsing/Parsers/Int.swift#L26-L49) defined to vend [`SubstringIntParser`](https://github.com/pointfreeco/swift-parsing/blob/346fbdcb793c4baa845ca44664ed20f719a7d83b/Sources/Parsing/Parsers/Int.swift#L129-L141), etc. I think we'd wanna do similar here.

Further, we probably want to add an `of inputType:` parameter to the `parser()` functions in order to disambiguate between `Substring.UTF8View`, `[UInt8]`, `Data`, etc., as we do with the `IntParser` here: https://github.com/pointfreeco/swift-parsing/blob/346fbdcb793c4baa845ca44664ed20f719a7d83b/Sources/Parsing/Parsers/Int.swift#L19

How does all this sound? Any questions or concerns?

stephencelis:
@lukeredpath Any chance you'll revisit this soon? Doing a lil spring cleaning and it stood out.

We'll definitely look to address this functionality in the future, especially as we explore invertible parsing, so if you don't have the time to get things across the finish line, no worries!

victorpimentel:
Hi @stephencelis! After watching your latest episode (awesome, as always) I had the same idea for the enum added in the episode, so before I found this issue I played a little bit with this idea:

https://github.com/victorpimentel/episode-code-samples/commit/19bbdb890f7622a170f25a987936787fbb512a5e

This seems like a very useful addition to the library, but I don't know if you are baking this in future episodes or if it would collide too much with the (supposedly) future builder additions.

If you give me the OK I'll be happy to take over this PR and apply your suggestions 😃

stephencelis:
@victorpimentel Very fun! We definitely think this would be a useful addition to the library, so if you'd like to PR it, or collaborate with @lukeredpath on a final version, we'd be happy to help get it merged!

I think your version already looks really great to move forward with, though! The one bike-shedding moment might be whether or not `.parser()` is descriptive enough. It might be, but we've also been considering `.allCasesParser()` or something similar.

lukeredpath:
@victorpimentel I'm pretty snowed under with work so please do feel free to take the ball and run with this. Take whatever you need from this PR if you need anything.

victorpimentel:
Nice! It has been a year since @lukeredpath opened this and he seems busy, but I'd like to give him a couple of days to answer if he wants to finish his work. Otherwise, I'd be happy to tackle this PR this week. 

About the _bike-shedding_, it's reasonable, I understand that it may not be obvious. However, just to be clear, wouldn't this be a bit out of sync with the rules of other existing parsers in the library? I'm referring to the point you made previously in https://github.com/pointfreeco/swift-parsing/pull/14#issuecomment-754106206

`Int.parser()`, `Double.parser()`, `Bool.parser()` and `Optional.parser()` are defined over sum types, and I believe that a case (no pun intended) can be made that these parsers are kind of specific cases of `CaseIterable.parser()`: they try to parse any of all possible cases for that sum type (I suppose `Bool` is the easiest to visualize, `Optional` works a bit differently and `Int` and `Double` have way more cases but still they are limited).

Are you afraid that it may lead to collisions with other future parsers? Right now I can't think of a `Role.parser()`, `Gender.parser()` or `Country.parser()` that can mean anything other than please try to parse the raw input into this enum (or raw representable type).

stephencelis:
> About the _bike-shedding_, it's reasonable, I understand that it may not be obvious. However, just to be clear, wouldn't this be a bit out of sync with the rules of other existing parsers in the library? I'm referring to the point you made previously in [#14 (comment)](https://github.com/pointfreeco/swift-parsing/pull/14#issuecomment-754106206)

I still think `.parser()` might be fine in this case, but I wanted to reopen the conversation. The issue is mainly when `.parser()` gets ambiguous because of overlapping protocol definitions. For example, what if we defined `.parser()` on `Decodable`, then if you conform a type to both `CaseIterable` and `Decodable` you start running into a problem.

(In this case, we think having an explicit `JSON()` parser is better, so the point is a bit moot, but hopefully gets to the heart of potential problems.)

victorpimentel:
Mmm, the type should then conform not only to CaseIterable but also to Decodable *and* RawRepresentable with different String literals, right? Sadly it can happen in the real world, although I'm not sure how good practice that is.

Anyway, whatever you prefer is fine to me :)

tgrapperon:
Are there already built-in parsers declared in protocol extensions in the library? (other than in extension of the `Parser` protocol!)

Maybe a safe route to avoid collisions would be to name these parsers methods `xXXParser()` where `XXX` is the protocol: `caseIterableParser()`, `rawRepresentableParser()` `decodableParser()`, etc. 
It is super easy for the user to alias a `.parser()` method to any of these methods if they want a leaner API.

stephencelis:
@lukeredpath Thanks again for your initial exploration of this. We'll be addressing it in an upcoming release, so going to close this for now!

PR #15, opened 2021-01-04
Optionally exclude newlines from Whitespace parser

ollitapa:
Some use-cases need newlines to be excluded from white space. For example if you want to parse Many() of parsers that contain trailing space with Newline() separator. Do you think this would be okay, or should I make a new parser like "Spaces" or "SpacesAndTabs"?

### DIFF
--- a/Sources/Parsing/Parsers/Whitespace.swift
+++ b/Sources/Parsing/Parsers/Whitespace.swift
@@ -5,17 +5,29 @@ where
   Input.SubSequence == Input,
   Input.Element == UTF8.CodeUnit
 {
+  public let includeNewlines: Bool
+
   @inlinable
-  public init() {}
+  public init(includeNewlines: Bool = true) {
+    self.includeNewlines = includeNewlines
+  }
 
   @inlinable
   public func parse(_ input: inout Input) -> Input? {
-    let output = input.prefix(while: { (byte: UTF8.CodeUnit) in
-      byte == .init(ascii: " ")
-        || byte == .init(ascii: "\n")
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @ollitapa! Thanks for the PR! We're still trying to figure out the balance for which parsers should ship with the library and which should maybe not (we don't want things to be _too_ overwhelming for users new to the library).

The `Whitespace` parser we have in the library is there as a nice, simple example of a custom parser, and it performs _slightly_) better than a `Prefix.init(while:)` parser that does the same thing. We're not sure if it even pulls its weight over `Prefix`, but we'll definitely keep thinking about this as the library develops, and consider introducing configuration as you do in this PR.

In the meantime, we think you should be able to get by with `Prefix { $0 == " " || $0 == "\t" }`, or if you want to really tune performance, you could even define your own `SpacesAndTabs` parser. How's that sound?

ollitapa:
Sure, no worries, I can roll with my own parser. I had the impression that Whitespace parser was preferred over Prefix, because it is in the same directory.

stephencelis:
> I had the impression that Whitespace parser was preferred over Prefix, because it is in the same directory.

We should probably document this 😅. The main reason we ship a dedicated `Whitespace` parser is for some otherwise tricky logic around newlines: it's a parser that operates on UTF-8 bytes, but it specifically handles the multibyte newline sequence of `\r\n`.

I'm going to close this PR for now, but if feel free to open a discussion if you'd like to continue the conversation and let others weigh in!

PR #17, opened 2021-01-13
Fix JSON parser

stephencelis:
This was on the branch with the JSON improvements but I forgot to push.

### DIFF
--- a/Sources/Parsing/Parser.swift
+++ b/Sources/Parsing/Parser.swift
@@ -66,9 +66,25 @@ extension Parser {
   where
     Input == Slice<UnsafeBufferPointer<UTF8.CodeUnit>>
   {
-    input.utf8.withContiguousStorageIfAvailable { input -> Output? in
-      var input = input[...]
-      return self.parse(&input)
-    }?.flatMap { $0 }
+    input.utf8
+      .withContiguousStorageIfAvailable { input -> Output? in
+        var input = input[...]
+        return self.parse(&input)
+      }?
+      .flatMap { $0 }
+  }
+
+  @inlinable
+  public func parse<C: Collection>(_ input: C) -> Output?
+  where
+    C.Element == UTF8.CodeUnit,
+    Input == Slice<UnsafeBufferPointer<C.Element>>
+  {
+    input
+      .withContiguousStorageIfAvailable { input -> Output? in
+        var input = input[...]
+        return self.parse(&input)
+      }?
+      .flatMap { $0 }
   }
 }
--- a/Sources/swift-parsing-benchmark/JSON.swift
+++ b/Sources/swift-parsing-benchmark/JSON.swift
@@ -5,10 +5,10 @@ import Parsing
 /*
  This benchmark shows how to create a naive JSON parser with combinators.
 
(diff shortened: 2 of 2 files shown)


Sources/Parsing/Parser.swift
```diff
@@ -66,9 +66,25 @@ extension Parser {
...
+    input.utf8
+      .withContiguousStorageIfAvailable { input -> Output? in
+        var input = input[...]
+        return self.parse(&input)
+      }?
+      .flatMap { $0 }
+  }
+
+  @inlinable
+  public func parse<C: Collection>(_ input: C) -> Output?
```
stephencelis, line 78: This could be pulled out to its own branch, but when benchmarking JSON I wanted to test `Data` as well and this makes it nicer.

Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -48,24 +48,26 @@ private let object = Skip(StartsWith("{".utf8))
         .take(stringLiteral)
         .skip(Whitespace())
         .skip(StartsWith(":".utf8))
-        .skip(Whitespace())
```
stephencelis, line 51: Removed some excess `Whitespace()` parsers (the parser that comes after is surrounded by them).

Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -48,24 +48,26 @@ private let object = Skip(StartsWith("{".utf8))
...
-      separator: Skip(Whitespace()).skip(StartsWith(",".utf8)).skip(Whitespace())
-    )
+      into: [:],
+      separator: StartsWith(",".utf8).skip(Whitespace())
+    ) { object, pair in
+      let (name, value) = pair
+      object[name] = value
+    }
   )
-  .map { JSON.object(Dictionary(uniqueKeysWithValues: $0)) }
```
stephencelis, line 56: Figured a `Many(_:into)` made more sense than a `Many(_:).map`.

Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -48,24 +48,26 @@ private let object = Skip(StartsWith("{".utf8))
...
+  .skip(StartsWith("}".utf8))
```
stephencelis, line 59: We were missing this close brace...so the parser wasn't quite right.

Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -174,10 +176,11 @@ let jsonSuite = BenchmarkSuite(name: "JSON") { suite in
     }
   )
 
+  let dataInput = Data(input.utf8)
   var objectOutput: Any!
   suite.benchmark(
     name: "JSONSerialization",
-    run: { objectOutput = try JSONSerialization.jsonObject(with: Data(input.utf8), options: []) },
+    run: { objectOutput = try JSONSerialization.jsonObject(with: dataInput, options: []) },
```
stephencelis, line 183: The benchmark wasn't "fair" to JSONSerialization since it was insuring the `Data.init` cost.

PR #21, opened 2021-02-08
Fix in documentation

tevelee:
In both examples, "Hello,World" is parsed the same way, therefore the remainders should be equivalent.


stephencelis (approved):
Ah yes, good catch, thanks!

PR #24, opened 2021-02-19
Add Take parsers up to 11

ryanbooker:
This may not be advisable. Chaining more that 6 parsers results in a noticeable increase in compile time. At 11 parsers it can be around 20 seconds. At around 16 parsers the compiler gives up. I've written the tests in a way that reduces compile times.

### DIFF
--- a/Sources/Parsing/Parsers/Take.swift
+++ b/Sources/Parsing/Parsers/Take.swift
@@ -66,6 +66,84 @@ extension Parser {
   where P: Parser, P.Input == Input, Output == (A, B, C, D) {
     .init(self, parser)
   }
+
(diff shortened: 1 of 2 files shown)


ryanbooker:
I noticed on my M1 Mac that testFloat() fails on:
```
    input = "1234567890123456789012345678901234567890 Hello"[...].utf8
    XCTAssertEqual(nil, parser.parse(&input))
    XCTAssertEqual("1234567890123456789012345678901234567890 Hello", String(input))
```
Parsing consumes the number, returning: `Optional(inf)`.

stephencelis (approved):
Thanks for taking the time to PR!

PR #25, opened 2021-02-19
Fix float overflow on utf8 parsing

ryanbooker:
I couldn't see a way to print the raw value of the Float80 to string—which prompted a feature enhancement idea: supporting scientific notation.

### DIFF
--- a/Sources/Parsing/Parsers/Double.swift
+++ b/Sources/Parsing/Parsers/Double.swift
@@ -128,7 +128,8 @@ extension Parsers {
       let original = input
       guard
         let s = input.parseFloat(),
-        let n = Double(String(decoding: s, as: UTF8.self))
+        let n = Double(String(decoding: s, as: UTF8.self)),
+        n <= Double.greatestFiniteMagnitude
       else {
         input = original
         return nil
@@ -163,7 +164,8 @@ extension Parsers {
       let original = input
       guard
         let s = input.parseFloat(),
-        let n = Float(String(decoding: s, as: UTF8.self))
+        let n = Float(String(decoding: s, as: UTF8.self)),
+        n <= Float.greatestFiniteMagnitude
       else {
         input = original
         return nil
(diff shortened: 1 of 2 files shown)


stephencelis:
Hi @ryanbooker, sorry for not following up sooner and thanks for PRing!

Would you mind walking me through this change a bit more? I checked out your branch and commented out the `n <= .greatestFiniteMagnitude` checks and the tests you added still seem to pass, so I'm just a bit puzzled as to what this PR changes 😅

ryanbooker:
Hi @stephencelis. Thanks for the reply. No worries at all. I neglected to mention this is related to issue: https://github.com/pointfreeco/swift-parsing/issues/26

It's only a problem in the Xcode 12.5 beta. Hopefully it will be fixed by release… it seems like an Apple regression.

ryanbooker:
Hi @stephencelis, I just received an update from Apple. This is new intended behaviour.

When parsing strings containing number values too large for the type, `Float` etc, now return `inf` rather than `nil`, so some version of this change will definitely be needed when Xcode 12.5 is released.

stephencelis:
Hm...I wonder if we should try to match Apple's behavior or not. My initial thought it it's probably better to diverge from this new behavior, but we'll chat about it and think things through.

Thanks for the update!

stephencelis:
@ryanbooker After thinking about this some more, we decided that it's probably best to embrace Apple's defaults here, even as they change, so we're going to keep things as they are.

We think that if you want to fail on infinity there are a couple options:

- Inspect the parser's output and fail if it's infinity
- Copy and paste the float parsers you use and modify the behavior in the project (if you want things to more immediately fail)

Thanks for bringing this to our attention, though!

ryanbooker:
No worries. My fear was a general one really. Relying on the old behaviour means everyone's code is now going to compile, but probably crash at run time. Apple did say they will update the release notes to actual mention this. :)

PR #31, opened 2021-05-12
Add FirstMatchOf parser

tgrapperon:
A parser that consumes elements of a collection one by one until a given parser `parser` succeeds.
In case of success, the prefix is discarded and the the output of `parser` is returned.

This parser is especially useful to extract some structured data inside an arbitrary collection, like a piece of source code for example.

### DIFF
--- a/Sources/Parsing/Parsers/FirstMatchOf.swift
+++ b/Sources/Parsing/Parsers/FirstMatchOf.swift
@@ -29,5 +29,5 @@ public struct FirstMatchOf<Input, P>: Parser
 }
 
 public extension Parsers {
-  typealias FirstMatch = Parsing.FirstMatchOf // NB: Convenience type alias for discovery
+  typealias FirstMatchOf = Parsing.FirstMatchOf // NB: Convenience type alias for discovery
 }
--- a/Sources/Parsing/Parsers/FirstMatchOf.swift
+++ b/Sources/Parsing/Parsers/FirstMatchOf.swift
@@ -1,0 +1,33 @@
+/// A parser that consumes elements of a collection one by one until and if a given parser `parser` succeeds.
+/// In case of success, the prefix is discarded and the the output of `parser` is returned.
+public struct FirstMatchOf<Input, P>: Parser
+  where
+  Input: Collection,
+  Input.SubSequence == Input,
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @tgrapperon! Thanks for taking the time to PR. This parser definitely is useful and we've played with similar functionality in the past. The one thing we're a little bit worried about is that use of such a parser could lead to performance issues. In particular:

```swift
Skip(FirstMatchOf(StartsWith("Hello".utf8)))

// vs.

Skip(PrefixUpThrough("Hello".utf8))
```

These should be functionally equivalent, but the former will run the parser against _every_ byte, whereas the latter will do a direct search on the input. I haven't benchmarked it, but I believe this _could_ be substantially slower. I'm wondering if you'd like to explore these performance issues a bit and think over the caveats. We could also put this PR on ice and open a discussion for now, to see if anyone else has thoughts on how to best accommodate the power of such a parser without accidentally introducing performance issues.

tgrapperon:
Yes @stephencelis, `FirstMatchOf` is indeed 9x slower in this specific example, benchmarked like you do for `PrefixUpTo`:
```
name                                         time          std        iterations
--------------------------------------------------------------------------------
PrefixThrough.Skip(PrefixThrough)             34381.000 ns ±  16.52 %      37342
PrefixThrough.Skip(FirstMatchOf(StartsWith)) 294248.000 ns ±   9.46 %       4649
```

Because the library is protocol-based, we can spot suboptimal configurations and I guess it's possible to automatically wrap `PrefixThrough` when this configuration is met. It is also very possible that other configurations are leading to other rules, including other parsers than `FirstMatchOf`. Please let me know if you're aware of such configurations.
Now that I've setup a benchmark suite, I'll investigate in this direction if you feel there's still life in it.

I initially implemented it to handle a collection which wasn't a `String` (or its derivative), and it was very difficult to instantiate a value for `PrefixUpTo` as I didn't owned the type. But as you said, it is definitely useful, and maybe faster than other solutions when it's the right tool, but it's not a silver bullet (yet).

I'll close the PR for now and reopen it if I'm achieving decent performance. I'll also open a discussion to gather optimization rules as may be a generalizable to other parsers.

PR #32, opened 2021-05-12
Fix testFloat() test

tgrapperon:
On my machine running Xcode 12.5, the test failed because it parsed `inf` instead of `nil`. I can't find in which version of Swift this behavior has changed, and I can't fence the code accordingly, but I guess this workaround still tests what matters.
I can happily close this PR if a better solution is found.


stephencelis (approved):
Thanks! This makes sense given the current behavior.

PR #34, opened 2021-05-12
Update format.yml

stephencelis:
This is broken everywhere we use the old swift-format action, so will PR whenever another project's `main` is broken.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -11,8 +11,10 @@ jobs:
     runs-on: macOS-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/format.yml
```diff
@@ -11,8 +11,10 @@ jobs:
     runs-on: macOS-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Tap
+        run: brew tap pointfreeco/formulae
```
stephencelis, line 15: I ended up forking (already using this on swift-composable-architecture after the last PR failed to run properly when merged into `main`).

PR #35, opened 2021-05-16
Add DelimitedBy

tgrapperon:
This parser matches content enclosed between pairs of delimiters. I feel the use-cases are general enough to justify a PR.
It is helpful when nesting of delimiters is unknown, as it will only match the top level pair. A maximum level of nesting can also be specified.
```
name                                       time          std        iterations
------------------------------------------------------------------------------
DelimitedBy.1-Level-DelimitedBy            166841.000 ns ±  17.58 %       7148
DelimitedBy.1-Level-DelimitedBy-MaxLevel-1  79969.000 ns ±  18.54 %      16657
DelimitedBy.1-Level-Composite               30869.000 ns ±  22.09 %      41878

DelimitedBy.SameDelimiter-DelimitedBy       18448.000 ns ±  22.99 %      70959
DelimitedBy.SameDelimiter-Composite         30881.000 ns ±  21.26 %      43653

DelimitedBy.Nested-DelimitedBy             167025.000 ns ±  15.00 %       7884
DelimitedBy.Nested-Ad-Hoc-Composite         40625.000 ns ±  17.86 %      31736
```
A benchmark suite is included. A 1-level `DelimitedBy` parser performs 2x slower than an adhoc parser, but with the guarantee that the result is enclosed in a matching pair (and if `maxLevel=1`, free of delimiters). If the level of nesting is not specified, the `DelimitedBy` parser performs 6x slower, with the same guarantee.

If the delimiters are the same, `DelimitedBy` performs 30% faster than an adhoc solution.

In case of nesting, `DelimitedBy` always takes the same time, but adhoc solutions will be more complex and slower as the supported level of nesting rises. It performs for example 4x slower than an adhoc parser made for matching with one inner pair exactly "`(xx(xx)x)`". I didn't implement branches for `(x(xx)x(xx)(x)...)` nor `(xxx)`. For deeper nesting, the combinatorics should immediately plays in favor of `DelimitedBy`, if not performance-wise, maintenance-wise.

Please let me know if you see room for improvement.

### DIFF
--- a/Sources/Parsing/Parsers/DelimitedBy.swift
+++ b/Sources/Parsing/Parsers/DelimitedBy.swift
@@ -1,0 +1,159 @@
+/// A parser that returns a subsequence by matching pairs of delimiters.
+/// The data must begin with `start`, and each successful match for `start` or `end` respectively open and close a matchin pair.
+/// The parser returns a subsequence of `Input` when it matches `end` with the initial `start`, closing the first level pair.
+/// In case of success, the output doesn't include the delimiters, and the last `end` portion of the input is consumed.
+/// This parser allows to parse arbitrary nested structures:
+///
+///     let delimitedParser = DelimitedBy<Substring>("(", ")")
+///
+///     var input = "(Hello,(world))!"[...]
+///     delimitedParser.parse(&input) // "Hello,(world)"
+///     input // "!"
+///
+/// If the input exceed the maximum level of nesting allowed, parsing will fail.
+/// By default, this maximum level is unlimited (`maxLevel: 0`)
+///
+/// - Warning: If the input is known to not have nested pairs of delimiters, an adhoc parser in the form of:
+///
+///     let parser = Parsers
+///     .Skip(StartsWith(start, by: areEquivalent))
+///     .take(PrefixUpTo(end, by: areEquivalent))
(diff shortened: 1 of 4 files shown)


stephencelis:
Hey @tgrapperon! Thanks for the PR 😄 This definitely looks useful, we just haven't had time to dive into things yet. We'll try to soon!

stephencelis:
Hey @tgrapperon! Thanks again for taking the time to PR! Brandon and I finally discussed today and while we think it's a cool parser, we think it may be a little too specialized for library inclusion at this time. It definitely solves a problem, but we're not quite ready for the library itself to commit to solving this particular problem. We want to keep the parser space that comes with the library lean for the moment, and super generalized, especially as we plan on layering on invertible parsing and other things soon, and so each additional parser we add is additional surface area to consider as the library evolves.

I think there are a few ways to proceed in the meantime:

1. I know we say it a lot...but plug-ins! 😄 We've been trying to ensure our libraries are extensible, which means extension code can ship outside the main library, and we think parsers like this are a perfect example of that.
2. You could share this code in a more visible place in the Discussions tab, so that others can find it.

I'm going to close the PR for now. When the library stabilizes in the future I think we can start to think more about what the requirements are for a parser to be included in the main library or not, and something like this may then fit the bill!

PR #37, opened 2021-05-27
Int parser fix on just "-" or "+"

mortigenus:
Fix for an issue discussed in https://github.com/pointfreeco/swift-parsing/discussions/36

Int parser should fail if a sign (-/+) is not followed by a digit.

Tests for Double and Float parsers are also added to prevent a possible introduction of this issue there in the future.

Note: I wanted to use `guard` for a check but both following examples didn't feel as straightforward as `if`:
```swift
guard !(parsedSign && length == 1) 
else { return nil }
```
```swift
guard !parsedSign || length != 1
else { return nil }
```

### DIFF
--- a/Sources/Parsing/Parsers/Int.swift
+++ b/Sources/Parsing/Parsers/Int.swift
@@ -121,9 +121,8 @@ extension Parsers {
         guard !overflow else { return nil }
         length += 1
       }
-      guard length > 0
+      guard length > (parsedSign ? 1 : 0)
       else { return nil }
-      if parsedSign && length == 1 { return nil }
       input.removeFirst(length)
       return output
     }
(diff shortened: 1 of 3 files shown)


Sources/Parsing/Parsers/Int.swift
```diff
@@ -119,6 +123,7 @@ extension Parsers {
       }
       guard length > 0
```
tgrapperon, line 124: I think `guard length > parsedSign ? 1 : 0` would allow to bypass the check of l126

mortigenus: That's a great idea, thanks!

stephencelis (approved):
Thanks for catching AND fixing! 😃

PR #38, opened 2021-06-22
Remove benchmark product

stephencelis:
Exposing the benchmark executable as a product makes its dependencies show up when depending on the package downstream. This means you'll see swift-argument-parser and swift-benchmark in your project when you depend on swift-parsing. By removing the product, these details should no longer be exposed downstream, and the benchmark still shows up just fine as an executable when working directly on swift-parsing.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -11,7 +11,7 @@
         }
       },
       {
-        "package": "Benchmark",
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):
nice

PR #45, opened 2021-07-02
DocC

stephencelis:
Upgrades the existing docs to use DocC. I think there are a lot of improvements we'll be able to make to the docs soon! May do a few PRs fleshing some stuff out.

### DIFF
--- a/Sources/Parsing/Parsers/StartsWith.swift
+++ b/Sources/Parsing/Parsers/StartsWith.swift
@@ -71,7 +71,6 @@ extension Parsers.StartsWith where Input.Element: Equatable {
     PossiblePrefix: Collection,
     PossiblePrefix.Element == Input.Element
   {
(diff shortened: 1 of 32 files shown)


Sources/Parsing/Parsers/StartsWith.swift
```diff
@@ -1,4 +1,24 @@
 /// A parser that parses a sequence of elements from its input.
+///
+/// This parser is named after `Sequence.starts(with:)`, and tests that the input it is parsing
+/// starts with a given prefix by calling this method under the hood.
```
mbrandonw, line 4: How about this to not further confuse this operator with the `prefix` operator?

```suggestion
/// starts with a given subsequence by calling this method under the hood.
```

mbrandonw (approved):
looks good! just one comment...

PR #47, opened 2021-08-18
Harness SE-0299 to improve parser discoverability and ergonomics

nikitamounier:
Hello there! I'd like to propose an important addition to the library.

 [SE-0299](https://github.com/apple/swift-evolution/blob/main/proposals/0299-extend-generic-static-member-lookup.md), called "Extending Static Member Lookup in Generic Contexts", offers a way to use static member dot-syntax in generic contexts, like so:
```swift
extension View {
  func buttonStyle<S>(_ style: S) -> some View where S: PrimitiveButtonStyle { /*...*/ }
}

struct BorderedButtonStyle: PrimitiveButtonStyle { /*...*/ }

// before:
Button(action: { viewStore.send(.increment) }) { /*...*/ }
  .buttonStyle(BorderedButtonStyle())

// after:
extension PrimitiveButtonStyle where Self == BorderedButtonStyle {
  static var bordered: Self { .init() }
}

Button(action: { viewStore.send(.increment) }) { /*...*/ }
  .buttonStyle(.bordered)
```

This feature was created due to the poor discoverability of these `...Style` structs. Now the code is much more readable, and Xcode can offer autocomplete suggestions.

This library unfortunately has the same problem – there are a lot of different parsers available, but their discoverability is very poor. There have been discussions about this already, with [one for example](https://github.com/pointfreeco/swift-parsing/discussions/4) proposing putting more parsers in the Parser namespace. 

Using this new language feature however, we can vastly improve discoverability and ergonomics without moving around any of the structs or even modifying them – this change is completely additive.

Here's an example of what this would look like:
```swift
let user = Int.parser()
  .skip(.startsWith(","))
  .take(.prefix { $0 != "," })
  .map { User(id: $0, name: String($1)) }
```

And here's an example of what adding one of the static members looks like. It looks a little different than the SwiftUI example because there are generics involved:
```swift
extension Parser {
  @inlinable
  public static func firstWhere<Input>(_ predicate: @escaping (Input.Element) -> Bool)
  -> Self where Self == FirstWhere<Input>, Input: Collection, Input.SubSequence == Input {
    return .init(predicate)
  }
}
```

I've implemented these static function extensions for:
- Always
- End
- Fail
- First
- FirstWhere
- Newline
- Prefix
- PrefixThrough
- PrefixUpTo
- Rest
- StartsWith
- Whitespace

These are parsers which make sense to use as a base parser (unlike Map, Skip, etc.) and they are also parsers which don’t wrap other parsers (such as Many, Lazy, etc.). I feel like these “higher-order parsers” should be created using their initialisers, similar to Publishers.MergeMany or Publishers.Zip2 - but I’d be happy to make static function extensions for those too if need be.

I feel like this addition would solve many of the problems people currently feel with discoverability and ergonomics while still keeping a familiar and idiomatic API.

### DIFF
--- a/Sources/Parsing/Parsers/Always.swift
+++ b/Sources/Parsing/Parsers/Always.swift
@@ -1,3 +1,4 @@
+import Foundation
 /// A parser that always succeeds by returning the provided value, but does not consume any of its
 /// input.
 ///
@@ -9,17 +10,17 @@
 /// ```
 public struct Always<Input, Output>: Parser {
   public let output: Output
-
+  
   @inlinable
   public init(_ output: Output) {
     self.output = output
   }
-
+  
   @inlinable
   public func parse(_ input: inout Input) -> Output? {
     self.output
   }
-
+  
   @inlinable
   public func map<NewOutput>(
     _ transform: @escaping (Output) -> NewOutput
@@ -30,4 +31,11 @@ public struct Always<Input, Output>: Parser {
 
 extension Parsers {
   public typealias Always = Parsing.Always  // NB: Convenience type alias for discovery
+}
+
+extension Parser {
+  @inlinable
+  public static func always<Input, Output>(_ output: Output) -> Self where Self == Always<Input, Output> {
+    return .init(output)
+  }
 }
--- a/Sources/Parsing/Parsers/Always.swift
+++ b/Sources/Parsing/Parsers/Always.swift
@@ -1,4 +1,3 @@
-import Foundation
 /// A parser that always succeeds by returning the provided value, but does not consume any of its
 /// input.
 ///
--- a/Sources/Parsing/Parsers/End.swift
+++ b/Sources/Parsing/Parsers/End.swift
@@ -7,7 +7,7 @@
 public struct End<Input>: Parser where Input: Collection {
   @inlinable
   public init() {}
-
+  
   @inlinable
   public func parse(_ input: inout Input) -> Void? {
     guard input.isEmpty else { return nil }
@@ -17,4 +17,11 @@ public struct End<Input>: Parser where Input: Collection {
 
 extension Parsers {
   public typealias End = Parsing.End  // NB: Convenience type alias for discovery
+}
+
+extension Parser {
+  @inlinable
+  public static func end<Input>() -> Self where Self == End<Input>, Input: Collection {
+    return .init()
+  }
 }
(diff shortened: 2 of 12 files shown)


stephencelis:
Hi @nikitamounier! Thanks for taking the time to look into this! It's something that's been on our radar, and we've been interested to see it in action, but we can't had time to consider the implications yet.

Have you tried going through the demo parsers in the benchmarks target and playground and update them to see what happens? It'd be nice to get a feel for the impact on some "real world" code in the repo.

Without seeing some parsers translated over, there are a couple things that I'm currently wondering about:

  - While button styles are always abbreviated inside a `buttonStyle` modifier, these parsers often appear "in the open" and at the start of chains (e.g., not just inside `skip` and `take` modifiers). This would seem to mean that they can't always be dot-abbreviated in this new syntax.

  - I think generics may make things even more complicated. A lot of the time when defining the smallest parsers in a larger parser, the generics must be explicitly spelled out. For example in our date parser demo:

    https://github.com/pointfreeco/swift-parsing/blob/625c1f23eb4902b5742a94e7d7a47d068d2bf68d/Sources/swift-parsing-benchmark/Date.swift#L30-L41

    Since `Parser` is a protocol with associated types and not a type with generics, is it even possible to use `.prefix` here? Or do we have to drop back down to capital `Prefix`, etc.? And if capital `Prefix` and friends _are_ still occassionally required, does this mean supporting 2 different syntaxes to do the same thing?

Do you have any clarity into these issues and ideas for potential workarounds?

nikitamounier:
As I've been updating the demo parsers and benchmarks, I've come to the conclusion that this change is a bit ahead of its time. There are a couple essential missing features in the language, including protocol metatype extensions, and more solid generics type inference on implicit static members. 

Taking a line from your code sample above:
```swift
private let timeHour = Prefix<Input>(2).pipe(Int.parser().skip(.end()))
```
The type-checker completely freaks out, saying:
> Cannot infer contextual base in reference to member 'end'

This is probably due to the generics involved – problem is, you can't specify generics in an abbreviated static function:
```swift
.end<String>() // some errors about binary operator '<'
```
Even worse, you can't do this:
```swift
Parser<String>.end()
```
because of the lack of protocol metatype extensions.

It's a shame, I think it would've greatly improved the ergonomics of the library, but it just isn't possible today.

stephencelis:
Given your conclusion I'm going to close this for now, but thanks again for digging deep into it. There definitely seems to be some promise here in the future, but it looks like we have to wait for the language to catch up. 😞

PR #48, opened 2021-08-30
Update FirstWhere.swift

morstin:
Dear all,

I have been trying to use FirstWhere with Substring.UTF8View. I find it difficult that FirstWhere returns a Collection.Element - in my case a Substring.UTF8View.Element. I did not find a way to convert this Element back to Substring.UTF8View. It seems to me that the only way to use the Substring.UTF8View.Element is to create a new String.

Assuming this is correct, I wondered if it would not be easier if the parser would return Substring.UTF8View as lot of others parser do.

In my fork you see an implementation proposal.

Thank you very much for the great library and the great video series!
Morstin

### DIFF
--- a/Sources/Parsing/Parsers/FirstWhere.swift
+++ b/Sources/Parsing/Parsers/FirstWhere.swift
@@ -11,10 +11,11 @@ where
   }
 
   @inlinable
-  public func parse(_ input: inout Input) -> Input.Element? {
-    guard let first = input.first, self.predicate(first) else { return nil }
+  public func parse(_ input: inout Input) -> Input? {
+      let firstOfCollection = input.prefix(1)
+      guard let firstElement = input.first, self.predicate(firstElement) else { return nil }
     input.removeFirst()
-    return first
+      return firstOfCollection
   }
 }
 


stephencelis:
Hi @morstin! Thanks for the feedback and PR. We try to keep these APIs close to Swift's built-in Collection APIs, and there `Collection.first(where:)`, which returns an `Element` and not a collection of one elements. However, re-reading, we're not sure `FirstWhere` should even be a parser, since as written it diverges from `Collection.first(where:)`, since it does not consume all elements before the first match. We maybe should deprecate it for now before completely removing it in the future.

In your case, though, I think you could use the `Prefix` parser instead, which will both leave things in a subsequence and can be limited with a given length:

```swift
Prefix(1, while: predicate)
```

stephencelis:
We've gone ahead and deprecated `FirstWhere`. Please let us know if `Prefix` isn't working for you as an alternative!

morstin:
Thank you very much for this very quick feedback. I tried to use Prefix before but had other problems with it. I assumed that with the maxLength you could limit the maximum number of consumed elements. 

But the way it works is that the parser fails if maxLength is exceeded: 

- use predicate to get part of collection
- fail if length of found collection is less than minLength
- fail if length of found collection is more than maxLength

What I was searching for was:

- do not consume more than maxLength elements
- consume and return if predicate is true (inside maxLength)
- require minLength otherwise fail

More concrete:

- consume exactly one character 
- as long it is not “\n”
- fail if you cannot consume this one character fail

The adapted version of FirstWhere works fine for me and I can keep it apart from the library…

stephencelis:
> I tried to use Prefix before but had other problems with it. I assumed that with the maxLength you could limit the maximum number of consumed elements.
> 
> But the way it works is that the parser fails if maxLength is exceeded:

This definitely feels like a bug in behavior. I've opened #50 to explore the other behavior. Brandon and I will discuss soon!

morstin:
Just a small comment - I am not sure this is a bug. I think the problem arises out of the possible different Prefix you can create through the init:

Prefix with min or max length
Prefix with predicate
Prefix with predicate AND with min or max length

(1) The implemented behavior could be described as 
Get Prefix according to predicate and check if it fits length requirements otherwise fail

(2) Perhaps it would be interesting to have a different parser for
Take a defined length of the input, consume based on the predicate and then check than minLength is fulfilled

This second parser does sound more like something that should be combined out of smaller parts. So I think it is not a bug, but I think the description of the behavior could be more precise for Prefix with predicate and min or max length…

stephencelis:
@morstin We just merged #50, as well, which should fix some behavior around parsing past the `maxLength`. I'm going to close this PR since I think we now support what it was trying to accomplish with existing operators. Thanks for opening it, though! Helped us catch what we feel was a bug in existing behavior.

PR #51, opened 2021-09-18
Fix Windows support, expand GHA platform coverage

MaxDesiatov:
Global `cLocale` constant was broken on Windows and was actually unused, it made sense to remove it.

Additionally, I've expanded macOS Xcode version matrix to include 12.4, 12.5, and 13.0 for testing with different Swift versions. Now all Xcode version values are quoted, since YAML automatically parses `13.0` as a number not a string, converting it to plain `13`, which breaks the build.

For testing on Windows I'm `MaxDesiatov/swift-windows-action` which invokes `swift test` automatically, but let me know if you'd like that to be more explicit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,6 @@ jobs:
           Copy-Item "$env:SDKROOT\usr\share\winsdk.modulemap" -destination "$env:UniversalCRTSdkDir\Include\$env:UCRTVersion\um\module.modulemap"
 
       - name: Build
-        run: swift build
+        run: swift build --target Parsing
       - name: Run tests
         run: swift test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   macos_tests:
-    runs-on: macOS-latest
+    runs-on: macos-11
     strategy:
       matrix:
         xcode:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
       - '*'
 
 jobs:
-  tests:
+  macos_tests:
     runs-on: macOS-latest
     strategy:
       matrix:
@@ -28,7 +28,7 @@ jobs:
       - name: Run ${{ matrix.command }}
         run: make ${{ matrix.command }}
 
-  windows_build:
+  windows_tests:
     runs-on: windows-2019
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - main
   pull_request:
     branches:
-      - '*'
+      - "*"
 
 jobs:
   macos_tests:
@@ -27,6 +27,20 @@ jobs:
         run: system_profiler SPHardwareDataType
       - name: Run ${{ matrix.command }}
         run: make ${{ matrix.command }}
+
+  ubuntu_tests:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        run: swift build
+      - name: Run tests
+        run: swift test
 
   windows_tests:
     runs-on: windows-2019
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,10 @@ jobs:
     strategy:
       matrix:
         xcode:
-          - 11.7
-          - 12.5
+          - 11.7 # Latest Xcode with Swift 5.2
+          - 12.4 # Swift 5.3
+          - 12.5 # Swift 5.4
+          - 13.0 # Swift 5.5
         command:
           - test
           - benchmarks
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - main
   pull_request:
     branches:
-      - "*"
+      - '*'
 
 jobs:
   macos_tests:
@@ -15,7 +15,7 @@ jobs:
       matrix:
         xcode:
           - 11.7
-          - 12.3
+          - 12.5
         command:
           - test
           - benchmarks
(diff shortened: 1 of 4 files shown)


stephencelis:
> Global `cLocale` constant was broken on Windows and was actually unused, it made sense to remove it.

Good catch, thanks! That was a relic from when we used `strtod_l`, but that was causing problems, so we decided to manually parse doubles instead, which wasn't actually so bad from a performance perspective compared to `strtod_l`.

stephencelis (approved):
Thanks! This all looks great!

MaxDesiatov:
Would it be possible to tag a new release that includes this change anytime soon? This would allow code depending on Swift parsing to specify this version instead of the branch for Windows support, which I'd assume is much safer from the API breakage perspective.

stephencelis:
@MaxDesiatov Done! https://github.com/pointfreeco/swift-parsing/releases/tag/0.2.0

Are you using swift-parsing in an open source project? Would love to see :smile:

MaxDesiatov:
It's very early stage, not much to see there yet or to be shared widely, but I'll just leave it here https://github.com/MaxDesiatov/Kara

stephencelis:
Thanks for sharing! If you hit any stumbling blocks or develop any interesting parsers that would be useful to contribute back, please open a discussion here if you find the time.

MaxDesiatov:
One interesting approach I came up with is using a specific structure for line/column counting that I then utilize for diagnostics and LSP support:

```
public struct ParsingState: Equatable {
  let source: String
  var index: String.UTF8View.Index
  var column = 0
  var line = 0
}
```

defined here https://github.com/MaxDesiatov/Kara/blob/main/Sources/Syntax/Parsing/StatefulParser.swift

I didn't not come up with a better name for it. I then use it as an input for all other parsers:

```
struct StatefulParser<P: Parser>: Parser where P.Input == UTF8SubSequence {
  let inner: P

  func parse(_ state: inout ParsingState) -> SourceRange<P.Output>? {
    let startColumn = state.column
    let startLine = state.line

    var substring = state.source.utf8[state.index...]
    let initialCount = substring.count

    guard let output = inner.parse(&substring) else {
      return nil
    }

    state.index = substring.startIndex
    state.column += initialCount - substring.count

    return SourceRange(
      start: .init(column: startColumn, line: startLine),
      end: .init(
        column: state.column - 1,
        line: state.line
      ),
      content: output
    )
  }
}

extension Parser where Input == Substring.UTF8View {
  func stateful() -> StatefulParser<Self> {
    .init(inner: self)
  }
}
```

This only works for wrapping parsers that don't consume newlines at all. I have a few other helpers for that if you're interested.

MaxDesiatov:
Overall, I found Swift Parsing to be super flexible and composable, and I'm very glad that I've chosen it for my project. The consensus on the internet seem to be that programming language parsers should be hand-written, and parser combinators are not a good fit because of their lack of diagnostics customization. This is definitely not the case here. Given that `Parser` type can accept any arbitrary input, I can easily add diagnostics infrastructure on top of it and it's very composable and flexible.

Thanks for the fantastic library and for tagging the new release promptly! As usual, your work is amazing!

stephencelis:
Glad to hear it! We have big plans for the project and will hopefully give it more attention soon.

- Invertible parsing is probably the thing we've had the most experience with (getting a printer for free when you define a parser, which is helpful for things like URL request routing, but also you could potentially get a pretty printer or minifier with little work).
- Error handling is a little vaguer, but we want parsing to support throwing in the future, rather than rely on `nil`.
- Async/await support is vaguer still, but something we definitely want to figure out for parsing streams of bytes.

MaxDesiatov:
> * Invertible parsing is probably the thing we've had the most experience with (getting a printer for free when you define a parser, which is helpful for things like URL request routing, but also you could potentially get a pretty printer or minifier with little work).

Oh my, I would love to see this! I saw your printers work in the other branch, but didn't know that printers can be inferred from parser definitions. I'm definitely interested to know how performant it is though.

> * Error handling is a little vaguer, but we want parsing to support throwing in the future, rather than rely on `nil`.

I wish this would be opt-in only. Throwing a single error from the parsing process is not desirable in my situation. For good diagnostics I need to collect multiple errors if possible. Overall, I think that error handling should not be implemented on the parsing level at all. As in, if we fail to consume all input, yes that's an error. But ideally I want my programming language parsers to be flexible enough to consume as much as possible of the "broken" input. Then on the higher level, after parsing is finished, I can go through invalid syntax nodes and audit them as needed, producing multiple detailed error messages. I'm following the approach described here https://duriansoftware.com/joe/constructing-human-grade-parsers

> * Async/await support is vaguer still, but something we definitely want to figure out for parsing streams of bytes.

Not really relevant for me personally, and I'm not sure what the use case could be, unless some async operations like file/network IO are needed for a parser to proceed. But then why wouldn't I get that data before the parsing started and feed that data to a parser synchronously?

Another thing I found useful I recalled just now is debugging parsers. It's very similar to Combine's `print` operator, I just stick it on top of any parser I find behaving erratically to help me diagnose issues https://github.com/MaxDesiatov/Kara/blob/main/Sources/Syntax/Parsing/Debug.swift

stephencelis:
> I hope this would be opt-in only. Throwing only a single error from the parsing process is not desirable in my situation. For good diagnostics I need to collect multiple errors if possible. Overall, I think that error handling should not be implemented on the parsing level at all, at least for any non-trivial parser. As in, if we fail to consume all input, yes that's an error. But ideally I want my programming language parsers to be flexible enough to consume as much as possible of the "broken" input. Then on the higher level, after parsing is finished, I can go through invalid syntax nodes and audit them as needed, producing multiple detailed error messages. I'm following the approach described here https://duriansoftware.com/joe/constructing-human-grade-parsers

I don't think adding throws would change things drastically. I think your experience would be the same, but you'd catch upstream errors and accumulate them rather than `nil`-test. Introducing a dedicated parsing error type would come with more benefits, like a better default debugging experience and the ability to generate pretty errors pointing to where parsing failed and why.

> Not really relevant for me personally, and I'm not sure what the use case could be, unless some async operations like file/network IO are needed for a parser to proceed. But then why wouldn't I get that data before the parsing started and feed that data to a parser synchronously?

It would allow you to drastically reduce the memory footprint of the application. If an app wanted to parse gigs of data, for example.

> Another thing I found useful I recalled just now is debugging parsers. It's very similar to Combine's `print` operator, I just stick it on top of any parser I find behaving erratically to help me diagnose issues https://github.com/MaxDesiatov/Kara/blob/main/Sources/Syntax/Parsing/Debug.swift

Nice! Yeah we've defined a similar parser in a project.

MaxDesiatov:
> It would allow you to drastically reduce the memory footprint of the application. If an app wanted to parse gigs of data, for example.

Right, I initially thought you'd handle this by parsing asynchronous chunks independently, which could work for CSV, but now thinking about it, is obviously not gonna work for JSON/YAML or XML.

PR #61, opened 2021-11-06
Add `Parsers.UTF8ViewToSubstring`, `Substring`/`UTF8View` inference overloads

stephencelis:
While we have `Parsers.Pullback`, which makes it easy to `.pullback(\.utf8)` any `UTF8View`-based parsers to substring, this isn't quite as performant as working directly with the substring-UTF8View boundary, nor is it as discoverable.

This PR introduces `Parsers.UTF8ViewToSubstring`, which is equivalent to many of the ad hoc `Substring*Parser` wrappers previously defined in this library, like `SubstringBoolParser`, `SubstringDoubleParser`, `SubstringIntParser`, etc.

As such, this PR also deprecates those ad hoc parsers for operators that simply wrap the `UTF8View`-based parser in `FromUTF8View`.

Finally, this PR introduces some more initializer overloads for `Substring` and `UTF8View` specifically (we had some for `Substring already, but let's flesh out some more), which are some of the most common inputs many of the parsers in this library works with. These initializers should help cut down on some angle bracket intensity.


Parsing.playground/Contents.swift
```diff
@@ -4,7 +4,7 @@ struct Color {
   let red, green, blue: UInt8
 }
 
-let hexPrimary = Prefix<Substring>(2)
+let hexPrimary = Prefix(2)
   .compactMap { UInt8($0, radix: 16) }
```
stephencelis, line 8: An example of explicit generics no longer being necessary for common inputs where the type can be inferred by a `map` (or `compactMap`).

Sources/Parsing/Internal/Deprecations.swift
```diff
@@ -0,0 +1,51 @@
+// NB: Deprecated after 0.3.1:
```
stephencelis, line 1: Moved all deprecations to a file that can be "versioned" to some extent.

Sources/Parsing/Parsers/Bool.swift
```diff
@@ -19,24 +19,34 @@ extension Bool {
...
+  /// - Parameter inputType: The `Substring.UTF8View` type. This parameter is included to mirror the
+  ///   interface that parses any collection of UTF-8 code units.
+  /// - Returns: A parser that consumes a Boolean value from the beginning of a substring's UTF-8
+  ///   view.
+  @inlinable
+  public static func parser(
+    of inputType: Substring.UTF8View.Type = Substring.UTF8View.self
+  ) -> Parsers.BoolParser<Substring.UTF8View> {
+    .init()
+  }
```
stephencelis, line 36: An example of a new overload that favors `UTF8View` when encountering a `UInt8` byte. Cuts down on things like `Prefix<Substring.UTF8View>`, where `Prefix` will now do.

Sources/Parsing/Parsers/Bool.swift
```diff
@@ -19,24 +19,34 @@ extension Bool {
...
+  ///   interface that parses any collection of UTF-8 code units.
   /// - Returns: A parser that consumes a Boolean value from the beginning of a substring.
   @inlinable
   public static func parser(
     of inputType: Substring.Type = Substring.self
-  ) -> Parsers.SubstringBoolParser {
-    .init()
+  ) -> FromUTF8View<Parsers.BoolParser<Substring.UTF8View>> {
+    .init(.init())
   }
```
stephencelis, line 50: We already had many overloads for `Substring`, but instead of ad hoc parsers for each instance, we can now use a single `FromUTF8View` wrapper instead.

Sources/Parsing/Parsers/End.swift
```diff
@@ -8,6 +8,14 @@ public struct End<Input>: Parser where Input: Collection {
   @inlinable
   public init() {}
 
+  @_disfavoredOverload
+  @inlinable
+  public init() where Input == Substring {}
+
+  @_disfavoredOverload
+  @inlinable
+  public init() where Input == Substring.UTF8View {}
```
stephencelis, line 17: Parsers beyond `Bool.parser()` and `Int.parser()` can benefit from this default inference, as well.

Sources/swift-parsing-benchmark/ReadmeExample.swift
```diff
@@ -43,7 +43,7 @@ let readmeExampleSuite = BenchmarkSuite(name: "README Example") { suite in
   }
 
   do {
-    let user = Int.parser(of: Substring.UTF8View.self)
+    let user = Int.parser()
```
stephencelis, line 46: Another example of improved inference.

Sources/swift-parsing-benchmark/StringAbstractions.swift
```diff
@@ -23,15 +23,15 @@ let stringAbstractionsSuite = BenchmarkSuite(name: "String Abstractions") { suit
 
   suite.benchmark("Substring") {
     var input = input[...].utf8
-    let output = Many(Int.parser(), separator: StartsWith("é").utf8).parse(&input)
+    let output = Many(Int.parser(), separator: "é".utf8).parse(&input)
```
stephencelis, line 26: We had some tests still using `StartsWith` that I noticed, so did a lil unrelated cleanup.

mbrandonw (approved):

PR #62, opened 2021-11-16
InfixOperator

stephencelis:
Introduces an `InfixOperator` parser for parsing expressions separated by an operator that can combine the value of each expression. (See `chainl1` and `chainr1` in other parser combinator libraries.)


Sources/Parsing/Parsers/InfixOperator.swift
```diff
@@ -0,0 +1,60 @@
...
+{
+  public let `associativity`: Associativity
+  public let expression: Expression
+  public let `operator`: Operator
+
+  @inlinable
+  public init(
+    _ operator: Operator,
+    associativity: Associativity,
+    do expression: Expression
```
mbrandonw, line 16: how about `operand` instead of `do`?

mbrandonw (approved):

PR #63, opened 2021-11-17
Clean up benchmarks

stephencelis:
This work based on a discussion we had earlier. Our benchmarks were not quite real world in the sense that they worked with unsafe buffer pointers. While it might be a good idea to eke out every last bit of performance with an unsafe pointer, our benchmarks should maybe err on the side of typical library use instead. As a consequence, this PR also cleans up how these parsers are defined.

### DIFF
--- a/Sources/swift-parsing-benchmark/CSV/CSVBenchmarks.swift
+++ b/Sources/swift-parsing-benchmark/CSV/CSVBenchmarks.swift
@@ -9,26 +9,20 @@ import Parsing
 
 // MARK: - Parser
 
-private struct CSV: Parser {
-  func parse(_ input: inout Substring.UTF8View) -> [[String]]? {
-    let plainField = Prefix {
-      $0 != .init(ascii: ",") && $0 != .init(ascii: "\n")
-    }
-
-    let quotedField = "\"".utf8
-      .take(Prefix { $0 != .init(ascii: "\"") })
-      .skip("\"".utf8)
-
-    let field = quotedField.orElse(plainField)
-      .map { String(Substring($0)) }
-
-    let line = Many(field, separator: ",".utf8)
-
-    let csv = Many(line, separator: "\n".utf8)
-
-    return csv.parse(&input)
-  }
+private let plainField = Prefix {
+  $0 != .init(ascii: ",") && $0 != .init(ascii: "\n")
 }
+
+private let quotedField = "\"".utf8
+  .take(Prefix { $0 != .init(ascii: "\"") })
+  .skip("\"".utf8)
+
+private let field = quotedField.orElse(plainField)
+  .map { String(Substring($0)) }
+
+private let line = Many(field, separator: ",".utf8)
+
+private let csv = Many(line, separator: "\n".utf8)
 
 // MARK: - Ad hoc mutating methods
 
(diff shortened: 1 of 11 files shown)


stephencelis:
Notably many parsers perform about as well, or up to twice as slow. Other benchmarks seem to be just old (`String.split` performance appears to have improved in Swift since we first ran the benchmarks).

Sources/swift-parsing-benchmark/CSV/CSVBenchmarks.swift
```diff
@@ -2,58 +2,33 @@ import Benchmark
...
+//name                        time           std        iterations
+//----------------------------------------------------------------
+//CSV.Parser                  1884886.500 ns ±  13.14 %        694
+//CSV.Ad hoc mutating methods 1229682.000 ns ±  12.79 %       1069
 
 // MARK: - Parser
 
-private typealias Input = Slice<UnsafeBufferPointer<UTF8.CodeUnit>>
+private struct CSV: Parser {
+  func parse(_ input: inout Substring.UTF8View) -> [[String]]? {
```
stephencelis, line 13: A stylistic change that also slightly improves performance... I think with performance these things may always need to be tested on a case-by-case basis, as this kind of stylistic change doesn't always seem to improve performance (I tried with the date time parser and it slowed things down slightly). Still, it's nice to house the logic in a namespace...

tgrapperon:
Should all versions with `UnsafeBufferPointer` be excluded from the benchmarks completely? 
The benchmark suite is not only measuring the performance of the library, but also showcasing how various scenarios can be accomplished. I think the `UnsafeBufferPointer` variants serve a pedagogical purpose, not by their result, but by the way the source `String` can be casted/wrapped into another type that is more efficient to parse. I wouldn't know how to do it myself if I didn't had the benchmark as an example (and I don't think they're used outside benchmarks in the repository).
Maybe it would be worthwhile to keep one or two around for simple cases, aside their `SubString` variants (I haven't checked in the PR if you're removing all of them).

Sources/swift-parsing-benchmark/CSV/CSVBenchmarks.swift
```diff
@@ -2,58 +2,28 @@ import Benchmark
...
 
-private let quotedField = Skip(StartsWith<Input>("\"".utf8))
+private let quotedField = "\"".utf8
   .take(Prefix { $0 != .init(ascii: "\"") })
-  .skip(StartsWith("\"".utf8))
+  .skip("\"".utf8)
 
 private let field = quotedField.orElse(plainField)
   .map { String(decoding: $0, as: UTF8.self) }
+//  .map { String(Substring($0)) }
```
mbrandonw, line 22: should we drop this commented line?

mbrandonw:
> `String.split` performance appears to have improved in Swift since we first ran the benchmarks).

I guess we could print the version of Swift in the benchmarks. We already print the machine details.

stephencelis:
> Should all versions with `UnsafeBufferPointer` be excluded from the benchmarks completely?
> The benchmark suite is not only measuring the performance of the library, but also showcasing how various scenarios can be accomplished. I think the `UnsafeBufferPointer` variants serve a pedagogical purpose

@tgrapperon Definitely! This full move is temporary, but we definitely want to better document/teach various means of eking out performance in the future. Unsafe buffer pointers are often, but not always, a bit faster, and this speed comes at the cost of knowing it's safe to do so. I think in the future we will want to:

1. Add documentation or DocC tutorials covering the topic of performance more explicitly than what can be inferred from a benchmark
2. Selectively add _additional_ unsafe buffer pointer benchmarks where appropriate

PR #70, opened 2021-11-30
Fix parsing of floats with scientific notation

oskarek:
The parsing currently fails when the `"E"` is followed by a `"+"`, because the `"E"` is not dropped before checking for that character.


stephencelis (approved):
Ah this was bad when refactoring recently! Thanks for catching!

PR #73, opened 2021-12-03
Check if `Many`'s `Upstream` is consuming its `Collection` `Input`

tgrapperon:
Here is a possible solution to the problem discussed in #72.
When the `Upstream.Input` of a `Many` parser is a `Collection`, the library checks that `input` is consumed in order to avoid infinite loops when the upstream parser can succeed without consuming the input.

For performance reasons, it checks that the `input.startIndex` is evolving when `input` is parsed. I'm not 100% sure it doesn't have edge cases with reversed collections or things like that. There's maybe a better property to check.

I don't know if it can work with `Sequence`'s, using another property.

This unfortunately leads to some code duplication for the public API's in order to branch into the correct initializers.

I'm using an optional block because I suspect it is faster to check than to evaluate one block that always returns `true` when input is not a `Collection`. This should be benchmarked, I guess.

### DIFF
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -40,7 +40,7 @@ where
   public let updateAccumulatingResult: (inout Result, Upstream.Output) -> Void
   public let upstream: Upstream
   @usableFromInline
-  internal let upstreamDidConsumeInput: ((_ input: Upstream.Input, _ rest: Upstream.Input) -> Bool)?
+  internal let inputWasConsumed: ((_ input: Upstream.Input, _ rest: Upstream.Input) -> Bool)?
   
   /// Initializes a parser that attempts to run the given parser at least and at most the given
   /// number of times, accumulating the outputs into a result with a given closure.
@@ -68,7 +68,7 @@ where
     self.separator = separator
     self.updateAccumulatingResult = updateAccumulatingResult
     self.upstream = upstream
-    self.upstreamDidConsumeInput = nil
+    self.inputWasConsumed = nil
   }
   
   /// Initializes a parser that attempts to run the given parser at least and at most the given
@@ -97,7 +97,7 @@ where
     self.separator = separator
     self.updateAccumulatingResult = updateAccumulatingResult
     self.upstream = upstream
-    self.upstreamDidConsumeInput = { $0.startIndex != $1.startIndex }
+    self.inputWasConsumed = { $0.startIndex != $1.startIndex }
   }
 
   @inlinable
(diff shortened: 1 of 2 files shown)


iampatbrown:
@tgrapperon nice work! Do you think also checking the `endIndex` would cover the edge cases without too much overhead?
```swift
$0.startIndex != $1.startIndex || $0.endIndex != $1.endIndex
```

tgrapperon:
@iampatbrown I've been bitten a few times with `SubSequence` indices, so I'm not considering myself an expert when dealing with them, and I'm sure I can arrive to two opposite conclusions if I think about them long enough. In any case,  there are two cases:
- `.startIndex` is always what the parser will consume first, and there is no need for an additional check.
- The parser can get receive any index of the input collection, and checking the `endIndex` too will not be enough.

I think we're in the first case, but I haven't tested with exotic input sequences. I need to do more tests to be sure.

I'm not aware of any overhead when accessing `startIndex` and `endIndex`. `count` is however O(n) if the collection is not `RandomAccessCollection`s (O(1) if it is). If we're in the second case, using `count` for `RandomAccessCollection`'s could be an efficient workaround, but it will restrict the checks to such collections.

iampatbrown:
@tgrapperon that make sense. I think your implementation would probably work for me in most cases.

tgrapperon:
I don't think the current version of `swift-parsing` is made to parse in arbitrary directions or non-contiguous chunks. 
If I'm not mistaken, all built-in parsers are working on the "head" of the input. If not already far out of the scope of the project, handling other scenarios would require changes in many places most likely.

In the vast majority of cases, parsing occurs on `String` (or subsequences of it). As a consequence, built-in parsers will not directly accept `ReversedCollection<String>.SubSequence` as input for example (or I'm missing some `Parser` transform). The underlying data can be parsed of course, but it will need to be flattened to something with increasing indices, bringing it back to the "standard" configuration.

I'm not missing something, `Parser`'s that can cause issues seem very theoretical and should probably be explicitly defined by users well aware of what they're doing. I can imagine parsing columns of a matrix flattened by rows for example (like a bitmap). In this case, we would check and consume non-contiguous portions of the input (but even then, starting from `startIndex`!). If this kind of parsing is in the scope of the library, maybe the way a `Parser` can consume its input should be formalized in order to compose them with the correct behaviors.

In other words, and like @iampatbrown said, this PR seems to handle most of the cases as it is.

tgrapperon:
I've updated the expected results, so `Many` behaves in a coherent way. 
There's a tricky edge case I've annotated in the code.

Combined with #77, we should be able to express many configurations in a natural and predictable way.
Interestingly, `NonEmpty` from #77 can also be applied to `Many` itself, so the user can decide if, or where, it should fail.

tgrapperon:
Closed in favor of #74.

PR #74, opened 2021-12-03
Use memcmp to short-circuit Many when input isn't consumed

stephencelis:
Just exploring an alternative implementation to #73.

Some unanswered questions:

- [ ] Can we come up with good examples of when we would _not_ want this behavior? _I.e._ what's a parser that could be fed to `Many`, not consume input, and succeed?
- [ ] Should this be a configuration option, as is done in #73?

@tgrapperon @iampatbrown Thoughts?

### DIFF
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -74,40 +74,46 @@ where
   public func parse(_ input: inout Upstream.Input) -> Result? {
     let original = input
     var rest = input
-    var previous = input
+    #if DEBUG
+      var previous = input
+    #endif
     var result = self.initialResult
     var count = 0
     while
       count < self.maximum,
       let output = self.upstream.parse(&input)
     {
-      defer { previous = input }
+      #if DEBUG
+        defer { previous = input }
+      #endif
       count += 1
       self.updateAccumulatingResult(&result, output)
       rest = input
       if self.separator != nil, self.separator?.parse(&input) == nil {
         break
       }
-      if memcmp(&input, &previous, MemoryLayout<Upstream.Input>.size) == 0 {
-        var description = ""
-        debugPrint(output, terminator: "", to: &description)
-        breakpoint(
-          """
-          ---
-          A "Many" parser succeeded in parsing a value of "\(Upstream.Output.self)" \
-          (\(description)), but no input was consumed.
+      #if DEBUG
+        if memcmp(&input, &previous, MemoryLayout<Upstream.Input>.size) == 0 {
+          var description = ""
+          debugPrint(output, terminator: "", to: &description)
+          breakpoint(
+            """
+            ---
+            A "Many" parser succeeded in parsing a value of "\(Upstream.Output.self)" \
+            (\(description)), but no input was consumed.
 
-          This is considered a logic error that leads to an infinite loop, and is typically \
-          introduced by parsers that always succeed, even though they don't consume any input. \
-          This includes "Prefix" and "CharacterSet" parsers, which return an empty string when \
-          their predicate immediately fails.
+            This is considered a logic error that leads to an infinite loop, and is typically \
+            introduced by parsers that always succeed, even though they don't consume any input. \
+            This includes "Prefix" and "CharacterSet" parsers, which return an empty string when \
+            their predicate immediately fails.
 
-          To work around the problem, require that some input is consumed (for example, use \
-          "Prefix(minLength: 1)"), or introduce a "separator" parser to "Many".
-          ---
-          """
-        )
-      }
+            To work around the problem, require that some input is consumed (for example, use \
+            "Prefix(minLength: 1)"), or introduce a "separator" parser to "Many".
+            ---
+            """
+          )
+        }
+      #endif
     }
     guard count >= self.minimum else {
       input = original
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -1,10 +1,4 @@
-#if canImport(Darwin)
-  import Darwin.C
-#elseif canImport(Glibc)
-  import Glibc
-#elseif canImport(WinSDK)
-  import WinSDK
-#endif
+import Foundation
 
 /// A parser that attempts to run another parser as many times as specified, accumulating the result
 /// of the outputs.
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -107,7 +107,6 @@ where
           ---
           """
         )
-        break
       }
     }
     guard count >= self.minimum else {
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -89,6 +89,14 @@ where
         break
       }
       if memcmp(&input, &previous, MemoryLayout<Upstream.Input>.size) == 0 {
+        breakpoint(
+          """
+          Many succeeded in parsing an \(Upstream.Output), but no input was consumed.
+
+          This is considered a logic error that would lead to an infinite loop, which should never \
+          happen.
+          """
+        )
         break
       }
     }
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -2,6 +2,8 @@
   import Darwin.C
 #elseif canImport(Glibc)
   import Glibc
+#elseif canImport(WinSDK)
+  import WinSDK
 #endif
 
 /// A parser that attempts to run another parser as many times as specified, accumulating the result
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -89,12 +89,22 @@ where
         break
       }
       if memcmp(&input, &previous, MemoryLayout<Upstream.Input>.size) == 0 {
+        var description = ""
+        debugPrint(output, terminator: "", to: &description)
         breakpoint(
           """
-          Many succeeded in parsing an \(Upstream.Output), but no input was consumed.
+          ---
+          A "Many" parser succeeded in parsing a value of "\(Upstream.Output.self)" \
+          (\(description)), but no input was consumed.
 
-          This is considered a logic error that would lead to an infinite loop, which should never \
-          happen.
+          This is considered a logic error that leads to an infinite loop, and is typically \
+          introduced by parsers that always succeed, even though they don't consume any input. \
+          This includes "Prefix" and "CharacterSet" parsers, which return an empty string when \
+          their predicate immediately fails.
+
+          To work around the problem, require that some input is consumed (for example, use \
+          "Prefix(minLength: 1)"), or introduce a "separator" parser to "Many".
+          ---
           """
         )
         break
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -74,29 +74,29 @@ where
   public func parse(_ input: inout Upstream.Input) -> Result? {
     let original = input
     var rest = input
+    var previous = input
     var result = self.initialResult
     var count = 0
     while
       count < self.maximum,
-      let output = self.upstream.parse(&input),
-      memcmp(&input, &rest, MemoryLayout<Upstream.Input>.size) != 0
+      let output = self.upstream.parse(&input)
     {
+      defer { previous = input }
       count += 1
-      rest = input
       self.updateAccumulatingResult(&result, output)
+      rest = input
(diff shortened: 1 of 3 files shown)


Sources/Parsing/Parsers/Many.swift
```diff
@@ -74,7 +80,11 @@ where
...
     var count = 0
-    while count < self.maximum, let output = self.upstream.parse(&input) {
+    while
+      count < self.maximum,
+      let output = self.upstream.parse(&input),
+      memcmp(&input, &rest, MemoryLayout<Upstream.Input>.size) != 0
+    {
       count += 1
       rest = input
       self.updateAccumulatingResult(&result, output)
```
stephencelis, line 90: We could potentially `memcmp` on the `result` instead...but that might create copies...

iampatbrown: Would it be a problem if `updateAccumulatingResult` doesn't actually modify `result` but still consumes input? eg. you might ignore a subset of the parser's output?

stephencelis: That too!

tgrapperon:
Is `memcmp` reliable enough to detect if something hasn't effectively changed? Many parsers are temporarily consuming a part of the input before restoring it in some way if they fail. What if `(consumption·restoration) != id`
I can imagine collections with the same content, but different memory (for example, a `Stack` that counts its mutations, or a naive `Deque` backed with 2 arrays, or any `Collection` that host external metadata along its elements). `memcmp` will work if the user always restore a copy of the original `input`[^1], but we have no guarantee that it's indeed the case. If the user pops a value, then pushes it back in the hypothetical stack that counts, the memory will change, but the parsed input won't, and `Many` will vend the same values on the next iteration.

Of course, we're in exotic lands, and I'm possibly worried by non-existing issues. Otherwise, some additional constraints are maybe required on the types or arguments to construct safer (and still efficient) parsers.

[^1]: I'm not even sure about this, as there are many optimization routes. I don't know if all types are always stored atomically on the heap for example.

tgrapperon:
> Can we come up with good examples of when we would not want this behavior?

The only contrived example I'm imagining is trying to parse potentially empty components. For example, in a CSV with empty cells: "value1,value2,,,,". The upstream parser should still succeed at the end where it will not consume the string. But I'm not 100% sure it's an example of the issue, as `Many`'s separator will most likely consume the separator and effectively make progress. Maybe there's a configuration where the `separator` parser can also succeed without consuming, but I can't find any concrete example too.

> Should this be a configuration option, as is done in Check if Many's Upstream is consuming its Collection Input #73?

Before you suggested a generic approach in #72, I've experimented a little with an additional `trait` property of `Parser`. 
This value was an `UInt`-backed `OptionSet` (so it can be extended by users), and various composite `Parser`'s would have to combine the upstream traits into new ones for themselves. This approach was motivated by `AnyParser` that erases a lot of information. The objective was to allow non-generic information to pass through. 

For example, we define `canSucceedWithoutConsuming = Trait(rawValue: 1<<0)`. This property can flow (or not) through various level of composition. If this value is detected in `Many`'s upstream, some additional checks can be performed to avoid infinite loops. Because the `trait` property is an instance value, it can depend on the parser's configuration. For example, `Prefix` has `canSucceedWithoutConsuming` if `minLength==0`, but not if `minLength==1`.

There's maybe a whole algebra of traits that can be exploited to preserve the performance (or even take shortcuts when possible), using the same kind of internal parametrization I've realized in #73. I haven't pushed farther yet.

This kind of optimization can probably be realized to some extent at the "type" level, but I'm worried it will lead to an explosion of parser types: `NonEmptyPrefix`, `Prefix`, etc.

iampatbrown:
> * Can we come up with good examples of when we would _not_ want this behavior?

Possibly an IPv6 address parser. `2001:db8:0:0:0:0:2:1` can be represented as `2001:db8::2:1`
It might be better suited for a custom parser, however, it's not out of the question for someone to try implement using `Many`.

Tests/ParsingTests/ManyTests.swift
```diff
@@ -89,4 +89,21 @@ class ManyTests: XCTestCase {
     )
     XCTAssertEqual(Substring(input), "")
   }
+
+  func testFailureIfNoProgress() {
+     var input = " Hello, World!"[...]
+     XCTAssertEqual(
+       Many(CharacterSet.alphanumerics).parse(&input),
+       []
```
tgrapperon, line 97: @iampatbrown I think the question @stephencelis was asking about parsers that succeed without consuming was related to this point.

I started reworking #73 because my assumptions were wrong (I checked for `nil` instead of `[]`). 
In this case, `.alphanumerics` succeed in parsing the empty string `""`. Strictly speaking, the returned value should be `[""]`, without consuming the string. This is absolutely not what I would expect while using `Many`, but since `CharacterSet` cannot fail as a parser, and `Many` reports the successes of its upstream, I think this is what it should return. Returning `[]` implies that `.alphanumerics` succeeded 0 times, which is not the case.

In @iampatbrown example, I absolutely expect the ipv6 parser to return the empty components as empty strings, but in this test example, I would expect `[]`, as I would be trying to parse alpha-numeric chars and found a space to begin with.

This kind of ambiguity seems to happen whenever the upstream parser output is a collection and can succeed with an empty collection. This is furthermore problematic with `Many` itself when the empty collection results from parsing the input without consuming, which is what #73 and #74 are trying to handle. 

Maybe this should be formalized at a higher level. I initially commented on #72 because I was surprised that `CharacterSet` parser would never fail. I suspect many users would be in the same position. Maybe one can introduce a higher order `NonEmpty` parser to wrap a parser of collection into a failing parser when it parses an empty collection?. This behavior is achievable by mapping the results only a posteriori, but it would maybe be clearer for the user if it would happen at the parser type level, as the behavior of a `NonEmpty<CharacterSet>` parser feels unambiguous. @stephencelis, do you want me to experiment with this?

iampatbrown: Ah okay. I think I get what you're saying now. 
These examples are inconsistent
```swift
func testNonEmptyString() {
  let input = "123"[...]
  let upstream = CharacterSet.alphanumerics.parse(input)
  let many = Many(CharacterSet.alphanumerics).parse(input)
  XCTAssertEqual(upstream.output, many.output?.first)
} // Passed

func testEmptyString() {
  let input = ""[...]
  let upstream = CharacterSet.alphanumerics.parse(input)
  let many = Many(CharacterSet.alphanumerics).parse(input)
  XCTAssertEqual(upstream.output, many.output?.first) 
} // Failed
```

tgrapperon: I've proposed #77. Using `.nonEmpty` on `CharacterSet.alphanumerics`, one should expect:
```swift
Many(.alphanumerics).parse(" Hello, World!") == (output: [""], rest: " Hello, World!")
Many(.alphanumerics.nonEmpty).parse(" Hello, World!") == (output: [], rest: " Hello, World!")

Many(.alphanumerics).parse("123") == (output: ["123"], rest: "")
Many(.alphanumerics).parse("") == (output: [""], rest: "")
Many(.alphanumerics.nonEmpty).parse("") == (output: [], rest: "")

.alphanumerics.parse("123") == (output: "123", rest: "")
.alphanumerics.parse("") == (output: "", rest: "")
.alphanumerics.nonEmpty.parse("") == (output: nil, rest: "")

Many(.alphanumerics.nonEmpty).nonEmpty.parse("") == (output: nil, rest: "")
```

stephencelis:
> Possibly an IPv6 address parser. `2001:db8:0:0:0:0:2:1` can be represented as `2001:db8::2:1` It might be better suited for a custom parser, however, it's not out of the question for someone to try implement using `Many`.
> 
> More specifically `1::` is the same as `0001:0000:0000:0000:0000:0000:0000:0000`

Because a separator is involved/consumed, parsing still succeeds:

```swift
func testEmptyElement() {
  var input = "2001:db8::2:1"[...]
  XCTAssertEqual(
    Many(Prefix(while: \.isHexDigit), separator: ":").parse(&input),
    ["2001", "db8", "", "2", "1"]
  )
}
```

Tests/ParsingTests/ManyTests.swift
stephencelis: @tgrapperon Definitely feel free to explore! We currently have things like `atLeast`/`atMost`/`minimum`/`maximum` options on `Many`/`Prefix`, but perhaps the concept could be generalized.

tgrapperon:
Using the configurable implementation of #73, I've benchmarked a few configurations.
- `nil` -> progress predictate is nil, always pass through
- `constant true` -> progress predicate always return `true`
- `startIndex` -> progress predicate is using #73 base implementation on `startIndex`
- `memcmp` -> progress predicate similar to this PR.

I've made sure were hitting the `Collection` overloads by adding `fatalError()` in the other implementations.

Please note I'm benchmarking the current loop implementation of #73 which is a little different from this PR. But it allows to compare the overload of `memcmp` "apples to apples".
```
name                                          time           std        iterations
----------------------------------------------------------------------------------
CSV.Parser - nil                              2614351.000 ns ±   9.07 %        445
CSV.Parser - constant true                    2510804.000 ns ±   4.04 %        544
CSV.Parser - startIndex                       2546036.000 ns ±   4.86 %        521
CSV.Parser - memcmp                           2552885.000 ns ±   5.44 %        542

HTTP.HTTP - nil                                  9230.000 ns ±  22.16 %     145810
HTTP.HTTP - constant true                        9224.000 ns ±  21.32 %     145501
HTTP.HTTP - startIndex                           9183.000 ns ±  21.32 %     146906
HTTP.HTTP - memcmp                               9201.000 ns ±  21.87 %     145519

JSON.Parser - nil                               12193.000 ns ±  20.62 %     111649
JSON.Parser - constant true                     12039.000 ns ±  21.33 %     108968
JSON.Parser - startIndex                        12031.000 ns ±  21.86 %     110487
JSON.Parser - memcmp                            12002.000 ns ±  19.08 %     111469

String Abstractions.Substring - nil           1096568.000 ns ±   5.41 %       1241
String Abstractions.Substring - constant true 1149551.000 ns ±   5.27 %       1160
String Abstractions.Substring - startIndex    1160821.000 ns ±   5.88 %       1174
String Abstractions.Substring - memcmp        1222424.000 ns ±   6.69 %       1081

Xcode Logs.Parser - nil                       6330140.000 ns ±   3.16 %        218
Xcode Logs.Parser - constant true             6330029.000 ns ±   3.83 %        219
Xcode Logs.Parser - startIndex                6308441.000 ns ±   3.46 %        219
Xcode Logs.Parser - memcmp                    6376244.000 ns ±   3.32 %        217
```
In worst case scenarios, the overhead of `memcmp` seems to be around 3-4%, but is often closer to 1%.

This seems acceptable, but perhaps additional checks on `Many` or other parsers can be configured using `@Environment` like it is experimented in the [`skip-whitespace`](https://github.com/pointfreeco/swift-parsing/tree/skip-whitespace) branch, with some `unsafe` flag to disable checks to favor performance in controlled situations, or maybe something more granular?

Other tests seem unconclusive. For example, I still don't know if it's preferable to use an optional block or a constant block for the fallback predicate when no checks are performed (but this is more an issue of #73, as this PR isn't configurable yet).

I've published a [`many-checks-benchmarks`](https://github.com/tgrapperon/swift-parsing/tree/many-checks-benchmarks) branch with the benchmarks.

stephencelis:
@tgrapperon @iampatbrown After messing with this a bit more we're thinking that the "breakpoint" style of loudly announcing the problem is probably the way to go, rather than quietly returning `[]` or `[""]`. We've updated the PR accordingly and are eager to hear if you think this is the direction we should move toward.

tgrapperon:
@stephencelis I think it's an acceptable solution. Maybe the fact that `Prefix` and `CharacterSet` can succeed without consuming and are unsafe to use as it in `Many` should be mentioned as a warning in the documentation of these parsers (and maybe in `Many`'s documentation too).

With `CharacterSet` at least, the kind of input that can cause issues is not exotic at all, and will most likely be encountered by the user while debugging/testing, as soon as the input changes its `CharacterSet`. The breakpoint will provide enough information to understand what's going on.

In this library, we express the requirement to parse things only once by consuming them. This is not working when the thing is an empty part of the input. I've inserted some logic in #73 so it can parse an empty part from the same spot only once, but this implies defining the "spot", hence retraining the effect to `Collection`s. This is also maybe inducing contradictions and paradoxes down the line, and I furthermore don't know if there are real-world uses. As this PR will take precedence, I'll close #73.

In order to preserve performance while making them more resilient, Parsers would need to carry more information about their behavior and results. The library doesn't report yet why a parser failed for example. If this is implemented someday, maybe these mechanics can be exploited to sidetrack metadata about the parsing and to hook up specific behaviors at runtime in composite parsers like Many.

mbrandonw (approved):

PR #77, opened 2021-12-04
Add `NonEmpty` parser for parsers of `Collections`

tgrapperon:
This parser may be useful combined with the behaviors discussed in #73 and #74.

### DIFF
--- a/Sources/Parsing/Parsers/NonEmpty.swift
+++ b/Sources/Parsing/Parsers/NonEmpty.swift
@@ -1,0 +1,35 @@
+extension Parser where Output: Collection {
+  /// Transforms the receiver parser of a `Collection` into one that fails when the output is empty.
+  @inlinable
+  public var nonEmpty: Parsers.NonEmpty<Self> {
+    .init(upstream: self)
+  }
+}
+
+extension Parsers {
+  /// A parser that fails when the output of another parser of a `Collection` type is empty.
+  ///
+  /// You will not typically need to interact with this type directly. Instead you will usually use
+  /// the ``Parser/nonEmpty`` operation, which constructs this type.
+  public struct NonEmpty<Upstream>: Parser where Upstream: Parser, Upstream.Output: Collection {
+    public let upstream: Upstream
+
+    @inlinable
+    public init(upstream: Upstream) {
+      self.upstream = upstream
+    }
+
+    @inlinable
+    public func parse(_ input: inout Upstream.Input) -> Upstream.Output? {
+      let original = input
+      guard let parsed = upstream.parse(&input) else {
+        return nil
+      }
+      if parsed.isEmpty {
+        input = original
+        return nil
+      }
+      return parsed
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


tgrapperon:
A few questions:

- [ ] Should it be defined using `Filter` instead of an ad-hoc implementation?
- [ ] Are there benefits in having this non-empty trait encoded at the type level?

The implementation in terms of `Filter` is as simple as:
```swift
extension Parser where Output: Collection {
  @inlinable
  public var nonEmpty: Parsers.Filter<Self> {
    self.filter { !$0.isEmpty }
  }
}
```
I'm not sure having to maintain a separate implementation for `Empty<Upstream>` is worth it. I'm not foreseeing any shortcuts or behaviors having this encoded as a type could enable. Any idea?

stephencelis:
This is fun! We're also exploring some alternate ways of handling this kind of configuration, which should hopefully make it possible to do things like:

```swift
CharacterSet.alphanumerics.atLeast(1)
```

tgrapperon:
> `CharacterSet.alphanumerics.atLeast(1)`

I've experimented with this too, but I was worried about `count` being O(n) for non `RandomAccessCollection`. Maybe two conditional conformances should be used in this case:
- One with `Output: RandomAccessCollection`, without any concerns;
- One with `Output: Collection`, with maybe a breakpoint warning when hitting this with an argument `> 1` (as `== 1` can be achieved using `!isEmpty`), or drop `atLeast` for non `RandomAccessCollection`'s.

Do you want me to expand on the idea in this draft, or are you going to take care of it internally?

tgrapperon:
I don't know how to implement `atMost` in an efficient way. With a naive implementation, we may parse thousands of elements with the inner parser, when at most one or two may be required. That's a missed opportunity for early exit.

Maybe something that can traverse implicitly the parsers arborescence like an `Environment` would help. I've seen something that may ressemble to that in the `skip-whitespace` branch, so I guess I'll wait to see what shape it takes.

I feel `atMost` it's a less frequent requirement though.

randomeizer:
Hmm, interesting. Was playing with this a bit myself - I have a model which uses `NonEmpty` values, and they are a bit of a pain to parse into. Currently just doing some `prerequisite` checks and then forcing it to assume the `NonEmpty` will not be `nil` after construction.

stephencelis:
@tgrapperon Given that this PR's functionality is currently helpers over `.filter`, do you think it's OK to close it out till we can figure out a more efficient way of solving the problem?

stephencelis:
> If you're interested, this is my current implementation for `OneOrMore`, which is essentially a stripped-down version of `Many`:

Thanks for sharing! Just a quick note that we've moved `Many`'s `breakpoint` check to be a throwing error now (`memcmp` did not appear to have a measurable impact on performance):

https://github.com/pointfreeco/swift-parsing/blob/85e1c50d03999a2b53d30ddae4d0f2a3d872b335/Sources/Parsing/Parsers/Many.swift#L143-L148

PR #80, opened 2021-12-17
Add `Separated` parser

tgrapperon:
This parser stands between the `Many` parser which requires the downstream parser to be the same, and a sequence of `take.skip(separator).take.skip(separator)…` which forces to skip the separator between each component and can be quite taxing for the compiler.

With this parser, one can parses a CSV line with typed columns for example:
```swift
Separated3(
  Int.parser(),
  Date.parser(),
  Coordinates.parser(),
  separator: ","
) // -> (Int, Date, Coordinates)?
```
It can also help when the compiler finds a parser too complex to resolve its type.

Right now, the user needs to specify the number of inner parsers, but I guess this can be improved and polished once the result builders story is defined in the library.

What are your thoughts?


mbrandonw:
Hi @tgrapperon, thanks for taking the time to sketch this out!

It's definitely interesting, and we do agree that lots of `.take(...).skip(...)`s can get really noisy. However, the API doesn't seem quite right that the user has to count the number of parsers they are going to use in order to figure out which `SeparatedX` to use.

We think there's a better way that dovetails with some of the parser builder stuff we've been exploring. We hope to be releasing that stuff in the January of 2022 and so once get some of the foundations in place we can discuss further.

tgrapperon:
Thanks for the feedback @mbrandonw. I've experienced with an overloaded factory on a caseless `Separated` enum. It was nice at call site, with things like `Separated.by(",")(p1, p2, p3,…)` but a little too heavy for the compiler (which was the primary motivation for this parser), and screams for a `ResultBuilder`. Since I knew that you were working on them, I've opted to wait and see the shape they'll take, and I've removed the factory to draft this bare-bone PR in order to discuss their mere existence.

I'll complete this PR once you release `ParserBuilder`s then.

tgrapperon:
Hey @mbrandonw & @stephencelis! I've started experimenting with a `ParserBuilder` variant of `Separated` and this is not as direct as I thought! The target API would be similar to `Many` and look like:
```swift
Separated {
// …
} separator: {
// …
}
```
This parser would be particularly useful to push the effective number of accepted parsers in a `ParserBuilder` forward, as many builder's slots are often "wasted" parsing separators. For example, looking at the benchmarks:
```swift
private let fullDate = Parse { // 5 blocks
  dateFullyear
  "-".utf8
  dateMonth
  "-".utf8
  dateMday
}
// would become
private let fullDate = Separated { // 3 blocks, 4x less routes to check for the compiler.
  dateFullyear
  dateMonth
  dateMday
} separator: {
  "-".utf8
}
```

The initial objective was to avoid introducing a plethora of new types and overloads, or another builder. The idea is to catch a `Zip` parser and to splice its internal parsers with some `Separator` parser, to generate a new parser with the same `Input` and `Output` as the original.

I've tried several variants, but I'm either ending up with a `Separated` accepting thousands of `init` overloads, which can cause issues with code completion when declaring the parser or with thousands of overloads to some `Parse {}.separated{ … }` function.

At the end of the day, I think that a dedicated builder is required.

One possible route to optimize things would be, without resorting to `buildFinalResult`[^1], to split `Zip[OV]*` parsers into two components: one `ZipN` that acts as a logic-less Tuple Parser, and one `NonVoid` or `Compact[OV]*` parser that maps the results to a non-void tuple, from `OOVO` to `OOO` for example. This would allow defining a `@SeparatedBuilder` parser that also uses both parsers, but also inserts a `SeparatedN` parser that transforms the `ZipN` into a "separated" version (that is parsing one separator between each component, and returning the same result as `ZipN` if succeeding: `ZipN -> SeparatedN -> Compact`, with the same parser generic signature as a standard, nonseparated version. Because `SeparatedN` will automatically discard the separated voids, the output is homogenous to a non-separated ZipN, and the same `Compact[OV]*` should be able to be used.
Unfortunately, if the `@SeparatedBuilder` can be generic over the `Separator`, it won't be able to receive the `Separator`'s value at the builder level (because `@ResultBuilder` build methods are all static), so it will probably need to return some parser factory instead of a parser value, and everything will be assembled in `Separated`'s initializer.

The other approach is to define a dedicated set of `Separated[OV]*` blocks. I will add roughly the same amount of code `@ParserBuilder` adds, but `@ParserBuilder` will stay untouched.

Maybe there are other ways, but I'm not seeing them yet.

[^1]: As `buildFinalResult` are off-topic for now, the existing `@ParserBuilder` will directly build a block of `CompactOOVO<Zip4>` parser instead of a `ZipOOVO` one. It merely returns a composite parser instead of an "atomic" one, and this shouldn't change a lot the builder's complexity.

Should I go forward and try this configuration, or is splitting `Zip[OV]*` parsers is this something you don't want to happen?
Are there other potential/preferred solutions?

tgrapperon:
As the library is evolving very quickly, the underlying code for this PR would need many adaptations.
I'm closing it for now and I'll eventually get back with another implementation once things have settled.

stephencelis:
@tgrapperon For what it's worth we think proper error messages will help motivate this operator even more. We're actively exploring that right now!

PR #81, opened 2021-12-23
Remove a few more unneeded StartsWith parsers.

mbrandonw:


stephencelis (approved):
Surprised we didn't catch these earlier!

Sources/swift-parsing-benchmark/StringAbstractions.swift
```diff
@@ -23,7 +23,7 @@ let stringAbstractionsSuite = BenchmarkSuite(name: "String Abstractions") { suit
 
   suite.benchmark("Substring") {
     var input = input[...].utf8
-    let output = Many(Int.parser(), separator: StartsWith("é").utf8).parse(&input)
+    let output = Many(Int.parser(), separator: "é".utf8).parse(&input)
```
stephencelis, line 26: This one's a bit confusing (and I've made the mistake a few times), but this `.utf8` is turning a substring parser into a UTF-8 parser.

```suggestion
    let output = Many(Int.parser(), separator: StartsWith("é").utf8).parse(&input)
```

mbrandonw: ah right, guess we'll have to wait to clean this up for when we have the `FromSubstring` parsers or whatever we decide to call them.

PR #83, opened 2022-01-05
Cleanup: generic names, routing example

stephencelis:
A couple things I noticed.

- While `Upstream` works well in some contexts, in others I think we can be more specific.
- `PathEnd` in the routing example should probably _only_ check the path.

### DIFF
--- a/Sources/Parsing/Parsers/Many.swift
+++ b/Sources/Parsing/Parsers/Many.swift
@@ -29,49 +29,49 @@ import Foundation
 /// precondition(input == "")
 /// precondition(output == 6)
 /// ```
-public struct Many<Upstream, Result, Separator>: Parser
+public struct Many<Element, Result, Separator>: Parser
 where
-  Upstream: Parser,
+  Element: Parser,
   Separator: Parser,
-  Upstream.Input == Separator.Input
+  Element.Input == Separator.Input
 {
   public let initialResult: Result
   public let maximum: Int
   public let minimum: Int
   public let separator: Separator?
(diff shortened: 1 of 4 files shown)


Sources/Parsing/Parsers/Many.swift
```diff
@@ -29,49 +29,49 @@ import Foundation
 /// precondition(input == "")
 /// precondition(output == 6)
 /// ```
-public struct Many<Upstream, Result, Separator>: Parser
+public struct Many<Element, Result, Separator>: Parser
```
stephencelis, line 32: Other ideas: `Each` or `EachParser`?

Also...this change _is_ breaking.

Introducing both a deprecated type alias _and_ a deprecated property seemed a bit heavy, but let me know and I'll add them!

Sources/Parsing/Parsers/Optional.swift
```diff
@@ -13,21 +13,21 @@ extension Optional {
...
 
 extension Parsers {
-  /// A parser that parses `nil` when its upstream parser fails.
+  /// A parser that parses `nil` when its wrapped parser fails.
   ///
   /// You will not typically need to interact with this type directly. Instead you will usually use
   /// `Optional.parser(of:)`, which constructs this type.
-  public struct OptionalParser<Upstream>: Parser where Upstream: Parser {
-    public let upstream: Upstream
+  public struct OptionalParser<Wrapped>: Parser where Wrapped: Parser {
```
stephencelis, line 20: Another generic name change...

mbrandonw (approved):

PR #89, opened 2022-01-13
* Added `Not` and `Peek` parsers

randomeizer:
* Allows previewing the next input without consuming it.
* Added tests for the above.

### DIFF
--- a/Sources/Parsing/Parsers/Peek.swift
+++ b/Sources/Parsing/Parsers/Peek.swift
@@ -1,0 +1,39 @@
+/// Parses a `Void` result if the next input matches the provided `Upstream` ``Parser``,
+/// otherwise returns `nil`, in both cases leaving the input unchanged.
+///
+/// For example, identifiers (variables, functions, etc.) in Swift allow the first character to be
+/// a letter or underscore, but not a digit, but subsequent characters can be digits.
+/// Eg. `let a1 = "a1"` is valid, but `let 1a = "1a"` is not. This parser will check the input
+/// starts with a letter or underscore, and if it does, it will return the remainder of the input
+/// up to the first character that is not a letter, a digit, or an underscore.
+///
+/// ```swift
+/// let identifier = Peek(Prefix(1) { $0.isLetter || $0 == "_" })
+///     .take(Prefix { $0.isNumber || $0.isLetter || $0 == "_" })
+/// ```
+///
+/// Note that the `Peek` does not consume the first character, but will only pass if the `Prefix(1)...` parses.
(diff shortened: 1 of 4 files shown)


Sources/Parsing/Parsers/Peek.swift
```diff
@@ -0,0 +1,39 @@
...
+    /// Construct a ``Peek`` with the provided `Upstream` ``Parser``.
+    ///
+    /// - Parameter upstream: The ``Parser`` to check.
+    @inlinable
+    public init(_ upstream: Upstream) {
+        self.upstream = upstream
+    }
+    
+    @inlinable
+    public func parse(_ input: inout Upstream.Input) -> Void? {
```
stephencelis, line 31: I think we want peek to be able to inspect the value, so this should return `Upstream.Output?` instead.

stephencelis (commented):
Looks good other than that one change!

Sources/Parsing/Parsers/Peek.swift
randomeizer: Hmm, so I'm order to ignore the `Peek`, it would have to be `.skip`ped as well? 

I can see how it might be useful occasionally, but most of the time I'm just using it as a check, and the subsequent parsing operations are including the "peeked" value already.

Is it with having two different operators, one which returns the Output (`Peek` makes sense here) and one which is Void? Maybe `Match` or `Require` for the latter?

stephencelis: > Hmm, so in order to ignore the `Peek`, it would have to be `.skip`ped as well?

If we check how other libraries do it, they definitely return the output:

https://docs.rs/nom/latest/nom/combinator/fn.peek.html

So I think an explicit `Skip { Peek { … } }` for when you want to ignore output is just fine. You won't even need the `Skip` if the parser you're peeking is a void parser, like string literals.

> I can see how it might be useful occasionally, but most of the time I'm just using it as a check, and the subsequent parsing operations are including the "peeked" value already.
>
> Is it worth having two different operators, one which returns the `Output` (`Peek` makes sense here) and one which returns `Void`? Maybe `Match` or `Require` for the latter?

I'd probably err on the side of just including `Peek` for now, and if you want to start a discussion exploring `Skip { Peek }` vs. `Match`/`Require`, please do!

randomeizer: True, the upstream parser type is passed on. I'll tweak it shortly.

randomeizer:
`Peek` has been updated to return the `Upstream.Output` value when successful.

stephencelis:
Thanks!

iampatbrown:
Great work @randomeizer :)

PR #94, opened 2022-01-21
Improve `Many` performance

stephencelis:
This makes `Many.Separator` internally non-optional, which improves the performance of benchmarks using `Many`. I'm a lil puzzled, because I thought we made `Many.Separator` optional in the first place for performance reasons, but perhaps more recent versions of Swift have caused the inverse to be true. Or maybe we were just mistaken in the first place!


mbrandonw (approved):
noice

PR #97, opened 2022-01-22
* Added `Expect` and test cases.

randomeizer:
`Expect` checks the input matches the provided upstream `Parser`. If so, it returns `Void`, otherwise, `nil`. Very similar to `Peek`, but doesn't return the parsed result.

### DIFF
--- a/Sources/Parsing/Parsers/Expect.swift
+++ b/Sources/Parsing/Parsers/Expect.swift
@@ -1,0 +1,39 @@
+/// Parses the provided `Upstream` ``Parser``, returning `Void` if it was successful or `nil` if not,
+/// but does not consume any of the input.
+///
+/// For example, identifiers (variables, functions, etc.) in Swift allow the first character to be
+/// a letter or underscore, but not a digit, but subsequent characters can be digits.
+/// Eg. `let a1 = "a1"` is valid, but `let 1a = "1a"` is not. This parser will check the input
+/// starts with a letter or underscore, and if it does, it will return the remainder of the input
+/// up to the first character that is not a letter, a digit, or an underscore.
+///
+/// ```swift
+/// let identifier = Expect(Prefix(1) { $0.isLetter || $0 == "_" })
+///     .take(Prefix { $0.isNumber || $0.isLetter || $0 == "_" })
+/// ```
+///
+/// Note that the `Expect` does not consume the first character, but will only pass if the `Prefix(1)...` parses.
+/// The subsequent `.take(...)` will consume the first character, along with any subsequent characters
+/// that match the criteria.
+public struct Expect<Upstream>: Parser where Upstream: Parser {
+  /// The parser from which this parser checks is successful.
+  public let upstream: Upstream
+
+  /// Construct a ``Expect`` with the provided `Upstream` ``Parser``.
+  ///
+  /// - Parameter upstream: The ``Parser`` to check.
+  @inlinable
+  public init(_ upstream: Upstream) {
+    self.upstream = upstream
+  }
+
+  @inlinable
+  public func parse(_ input: inout Upstream.Input) -> Void? {
+    let original = input
+    if self.upstream.parse(&input) != nil {
+      input = original
+      return ()
+    }
+    return nil
+  }
+}
--- a/Tests/ParsingTests/ExpectTests.swift
+++ b/Tests/ParsingTests/ExpectTests.swift
@@ -1,0 +1,30 @@
+import Parsing
+import XCTest
+
+class ExpectTests: XCTestCase {
+  
+  func testExpectMatches() throws {
+    var input = "_foo1 = nil"[...]
+
+    let identifier = Expect(Prefix(1) { $0.isLetter || $0 == "_" })
+      .take(Prefix { $0.isNumber || $0.isLetter || $0 == "_" })
+
+    let result = identifier.parse(&input)
+
+    XCTAssertEqual(result, "_foo1")
+    XCTAssertEqual(input, " = nil"[...])
+  }
+
+  func testExpectFails() throws {
+    var input = "1foo = nil"[...]
+
+    let identifier = Expect(Prefix(1) { $0.isLetter || $0 == "_" })
+      .take(Prefix { $0.isNumber || $0.isLetter || $0 == "_" })
+
+    let result = identifier.parse(&input)
+
+    XCTAssertNil(result)
+    XCTAssertEqual(input, "1foo = nil"[...])
+  }
+
+}


stephencelis:
Hi @randomeizer! Thanks for taking the time to PR! And for being active in our discussions!

I definitely see this operator reading nicely! I think the main question we might ask is around the weight of adding surface area to the base library. Parsing can be a pretty intimidating thing to learn, so minimizing concepts at the start can go a long way. If anything, we'd like to constantly look to eliminate surface area where we can, or make it available progressively. We've been prepping some big releases soon and will be trying to do just this.

While we haven't formalized it anywhere, I can try to describe a few ways we evaluate whether a dedicated parser should ship in the base library:

1. Is it a universal operation and/or does it have a precedent elsewhere (do other parsing libraries ship with something similar)?
2. Is it not possible to easily recreate the functionality of this parser from existing units?
3. While it may be possible to easily recreate the functionality of this parser from existing units, would doing so be detrimental to performance?

In this case, `Expect` is equivalent to `Skip` + `Peek`:

```swift
Skip { Peek { ... } }
Expect { ... }
```

And while I agree it may read nicer, instead of adding another parser with the base library, I think it would make more sense for library authors to pave the way, kind of like how [RxSwift](https://github.com/ReactiveX/RxSwift) has [RxSwiftExt](https://github.com/RxSwiftCommunity/RxSwiftExt), or for a user that needs this combined functionality a lot in a project to define the operator themselves. You can even do it in a few lines if you don't want to introduce the ceremony of another whole type:

```swift
func Expect(@ParserBuilder _ build: () -> P) -> Skip<Peek<P>> {
  Skip { Peek { build() } }
}
```

Hopefully this thought process makes sense! We also welcome more discussion in #93. It's possible that others would benefit from this operation and would like to copy and paste it into their code bases or even create a plugin for it!

iampatbrown:
> I can try to describe a few ways we evaluate whether a dedicated parser should ship in the base library

This is some good info. Thanks.

> so minimizing concepts at the start can go a long way.

I think this is the main reason I'm hesitant about this addition. It's definitely a useful parser, it just feels more suitable as an extension at the moment.

randomeizer:
Yeah, `Skip` + `Peek` can achieve the same result. The main reason I pitched it it because in practice, I've used `Expect` more than `Peek` in my actual parsers. In fact, exclusively.

What parsing situations are more likely to use `Peek`?

haikusw:
@stephencelis Would it make sense to add something like this to the ReadMe?  Seems like very good information for potential contributors to see.

> While we haven't formalized it anywhere, I can try to describe a few ways we evaluate whether a dedicated parser should ship in the base library:
> 
> 1. Is it a universal operation and/or does it have a precedent elsewhere (do other parsing libraries ship with something similar)?
> 2. Is it not possible to easily recreate the functionality of this parser from existing units?
> 3. While it may be possible to easily recreate the functionality of this parser from existing units, would doing so be detrimental to performance?
> 4. Can this behavior be added as configuration to an existing operation?

stephencelis:
@haikusw Good idea! I think there probably a few places we could improve communication:

- A "Contributing" section of the README that directs folks to the GitHub discussions for "Show and Tell"-ing new parsers they come up with.
- A pinned discussion that better formulates the thought process behind adding new parsers to the library.

We'll try to get to this soon!

haikusw:
@stephencelis Sounds good.

I just ran across another similar chunk of overview thinking / philosophy in https://github.com/pointfreeco/swift-parsing/discussions/4#discussioncomment-236122

> @mayoff Thanks for starting the discussion! We're definitely open to refine the ergonomics of the library. As far as our mentality has been with what we've shipped so far, @mbrandonw pointed out that we've mostly tried to follow Combine's lead. This means establishing a few loose rules:
> 
> 1. If the parser is primarily vended from the method of another parser, the type should be nested in `Parsers`. (Like many publishers in Combine: `Publishers.Map`, `Publishers.FlatMap`, etc.)
> 2. If the parser is only meant to be used at the start of a chain, like `Prefix` and `StartsWith`, the type should be at the module level. (Like the `Just`, `Fail`, `Future`, `Deferred` Combine publishers, and more.)  A sub-rule is that these "free" parsers should try not to collide with Combine publishers. It's idealistic, of course, and we have one exception with the `Fail` parser (if you have other name ideas that don't collide, let us know!), but generally we've found appropriate alternatives (`Just`→`Always`, `Deferred`→`Lazy`, etc.).
> 3. If it's a parser of a specific type as output, it should be exposed as a static member on that type, like `Int.parser()`, `Bool.parser()`, etc. This is a _nod_ to Combine, which exposes publisher properties and methods on optionals, sequences, and URL sessions, but isn't quite the same.
> 
> One other thing we do for 2. and 3. is add a convenience type alias to them inside the `Parsers` namespace, so you can explore all of the publishers the library contains via autocomplete. (This is an experiment that we hope is helpful, but if it's the opposite we'll probably remove them.)
> 
> While we have established these rules, the reality is that they are loose and could be refined. For example:
> 
> * `Optional.parser(of: anotherParser)` could be a method, like `anotherParser.optional()`, but we found that "optional parser of another parser" reads better than "another parser, optional". This could probably be bike-shed to something just as nice or nicer, though? Like maybe `anotherParser.optionally()`, `anotherParser.orNil()`, etc.
> * I think Combine still nests some publishers that are meant to be used at the start of the chain, like `Publishers.MergeMany`, So maybe we should do the same with `OneOfMany` (especially since `Parsers.OneOf` via `orElse` should be preferred unless you're really given the opportunity to improve performance with `OneOfMany`, and maybe that opportunity is narrow enough that we should remove it from the core library entirely, and instead let performance-critical users to define it, or that kind of branching, from scratch?).
> * We should also maybe tighten up the number of module-"free" parsers defined when possible, like removing or nesting `OneOfMany`, and folding `PrefixThrough` and `PrefixUpTo` into the `Prefix` parser.
> * We also deviate a bit from Combine's practice of namespacing publishers like `DataTaskPublisher` inside `URLSession`, not sure if it makes sense to move `IntParser` into the `Int` type, etc.
> 
> If you have any specific ideas we'd love to hear them! We'd also love to hear of any specific instances in which having top-level parser types interferes with your workflow. Are you having a conflict in a particular code base because of our `Prefix` type, or our `StartsWith` type, etc.?

May be a few more similar things scattered about (I only looked at two) that could be brought together.

stephencelis:
Thanks for digging that up! A lot has evolved since then :smile:

- `StartsWith` is usually avoidable and one can use strings and UTF-8 views directly instead
- With parser builders we've migrated to incorporate some SwiftUI inspiration over Combine (though there's overlap, of course)
- `Optional.parser(of:)` is now `Optionally { ... }`
- `OneOfMany` _has_ been nested as the builder behind `for`-`in` loops in `OneOf { ... }`

We cover some design philosophy in the README, but we could go further. And there are more things to come as Swift evolves further!

stephencelis:
@randomeizer Is it okay to close this out for now since the operation can be recreated so easily in your own project? Should more folks request a similar compound operation in the future we can revisit things.

stephencelis:
@randomeizer Anecdotally, we've worked in parsing-printing for awhile and are polishing functionality for making many parsers into printers/formatters automatically, but one of the few parsers that is not printable is `Peek`, and so `Peek`-like functionality needs to be achieved by other means.

We're not sure if printability is even a concern of yours, but if so we'd be curious if there are other parsers out there more suited to the problem that can retain printability.

stephencelis:
@randomeizer We don't think so. We think most parsers that don't consume cannot reasonably print, since printing reverses the operation of parsing, and we couldn't find a reasonable way of reversing the operation of peeking.

That said, the parsers we've built have never used `Peek` or `Not`, so we're less familiar with them. There could be a reasonable way of making them printable, but we couldn't figure out a way that made sense to us, being less familiar with them.

randomeizer:
I would imagine that any parsers that consume unspecific content and discard it would be unable to reproduce it. For example, `Skip { Prefix { !$0.isNewline } }` would also be unprintable, right?

stephencelis:
@randomeizer You can always print skipped content by overriding it with a default. I'm not sure I follow your example, but you could skip whitespace with a default of a single space:

```swift
Skip { Prefix { $0.isWhitespace } }
  .printing(" ")
```

randomeizer:
My example was more that you're losing information - any non-newline would be lost. It could certainly be printed as `""` or whatever you like, but the data is gone.

I still haven't had a chance to dig into the printing  branch, but why can't you just do something like:

```swift
Not { "Foobar" }.printing("")
```

stephencelis:
> My example was more that you're losing information - any non-newline would be lost. It could certainly be printed as `""` or whatever you like, but the data is gone.

Yup! And that's a decision you have to make whenever you build a parser (and/or printer). You could preserve a lot of information, like Swift Syntax, and parse "trivia" to keep the exact whitespace information around, or you could forget about whitespace and instead omit or auto-format the printing of whitespace.

> I still haven't had a chance to dig into the printing branch, but why can't you just do something like:
> 
> ```swift
> Not { "Foobar" }.printing("")
> ```

I would need to see a more complex example of how you'd want to parse/print, but in losing the information, you would be able to print invalid things. _E.g._:

```swift
let uncommentedLine = Parse {
  Not { "//" }.printing("")
  Prefix { $0 != "\n" }
}

uncommentedLine.print("// hello from comment") // prints fine
```

We want to preserve a fuzzy kind of round-tripping as much as possible when building parser-printers.

randomeizer:
Does `print(...)` need to be able to `throw` then?

iampatbrown:
I've been using a `PrefixBy` parser which is similar to `Prefix` but takes `computeDistance:` instead of `while:`

```swift
public init(
  minLength: Int = 0,
  maxLength: Int? = nil,
  computeDistance: @escaping (Input, Output) throws -> Int
) {
  self.minLength = minLength
  self.maxLength = maxLength
  self.computeDistance = computeDistance
}
```

The `computeDistance:` closure can get a bit messy but it's helped create some useful parsers while maintaining printability.

```swift
var input = """
struct Blob {
 var name = "Blob"
}
"""[...]

let uncommentedLine = PrefixBy { input, output in
  guard let first = input.first, first != "\n" else { return 0 }
  guard !output.isEmpty || !input.starts(with: "//")
  else { throw ParsingError.failed(summary: "invalid prefix", at: input) }
  return 1
}

XCTAssertEqual("struct Blob {", try uncommentedLine.parse(&input))
XCTAssertEqual("struct Blob {", try uncommentedLine.print("struct Blob {"))
XCTAssertThrowsError(try uncommentedLine.print("// hello from comment")) 
```

stephencelis:
> Does `print(...)` need to be able to `throw` then?

Yup! It can and does on the printer branch.

randomeizer:
Since this conversation is on a closed PR and I've had a chance to look at the `printing-throws-2` branch, I've started a new discussion at #144.

randomeizer:
I'm happy to spin up a PR. While I'm not sure that `Expect` is the right name either, if you've ever used a stack or a queue, `Peek` does imply you'll get to see the result. I suppose the wrapped parser does get to preview it though.

stephencelis:
@randomeizer Yeah, I think even though the behavior is slightly different than what it is in nom and parsec and other libraries, we're not yet convinced that `Peek` in this form is a good parser to reach for. Maybe there are examples of this that we're not seeing, but it seems like the `Expect` behavior of `Peek` is what we think folks should use when building these parsers.

I think we're comfortable moving forward with a bit of divergence here till convinced otherwise, since `Not` is used as a kind of inverse `Peek`, and it has the nice property of:

```swift
Peek { P } == Not { Not { P } }
```

randomeizer:
No worries. Pushed #162 to implement this against `main`.

PR #98, opened 2022-01-24
Add character set parser tests

andrewjl:
- <s>Return nil when character set parser has no output</s>
- Add character set parsing tests

The rest of the test suite passes as well.

<s>Based on the [documentation](https://pointfreeco.github.io/swift-parsing/Parser/), my understanding is that a `CharacterSet`-based parser needs to return `nil` when the first character of its input isn't a member of the character set. </s>

I think there might be edge cases with non-ASCII characters that might be worth adding to tests, which I'm happy to do. Added some basic `CharacterSet` tests and tests for this specific change as a starter.

### DIFF
--- a/Tests/ParsingTests/CharacterSetTests.swift
+++ b/Tests/ParsingTests/CharacterSetTests.swift
@@ -1,0 +1,46 @@
+import Parsing
+import XCTest
+
+final class CharacterSetTests: XCTestCase {
+    func testDecimalDigits() {
+        let characterSet = CharacterSet.decimalDigits
+
+        var input = "42abc"[...]
+
+        let output = characterSet.parse(&input)
+        XCTAssertEqual(output, "42")
+        XCTAssertEqual(input, "abc")
+    }
+
+    func testAlphanumerics() {
+        let characterSet = CharacterSet.alphanumerics
+
+        var input = "42abc;."[...]
+
+        let output = characterSet.parse(&input)
+        XCTAssertEqual(output, "42abc")
+        XCTAssertEqual(input, ";.")
+    }
+
+    func testCustom() {
+        let characterSet = CharacterSet(charactersIn: "0123")
+
+        var input = "23456789abcdef"[...]
+
+        let output = characterSet.parse(&input)
+        XCTAssertEqual(output, "23")
+        XCTAssertEqual(input, "456789abcdef")
+    }
+
+    func testCompletion() throws {
+        let characterSet = CharacterSet(charactersIn: "123")
+
+        var input = "123456"[...]
+
+        let firstOutput = try XCTUnwrap(characterSet.parse(&input))
+        let secondOutput = try XCTUnwrap(characterSet.parse(&input))
+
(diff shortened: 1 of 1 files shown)


iampatbrown:
Hey @andrewjl,

I'm pretty sure this is the intended behaviour. This [comment](https://github.com/pointfreeco/swift-parsing/commit/d895babd90c3739ed41de53bb45ef873d56b1c37#commitcomment-61272825) from @stephencelis gives a bit more context:

> This parser is much like Prefix(while:), which also can't fail by predicate alone. If output is empty parsing should still succeed, and you can instead do a filter { !$0.isEmpty }/pipe(Prefix(atLeast: 1) to fail over.

Things may have changed though. Just wanted to give you a heads up.

andrewjl:
Hey @iampatbrown,

Thanks for chiming in! I read over the linked comment thread, which does make sense. 

I think if returning an empty string in this case is the intended behavior for `CharacterSet`, then the documentation for the `Parser` protocol should account for this. Perhaps by clarifying exactly what "parsing failure" means?


@stephencelis let us know your thoughts and I can close this PR if it's not the correct change. 🙂

stephencelis:
> Actually, I ran into a similar edge case to what happens with `Many` that was described in a follow-on [comment](https://github.com/pointfreeco/swift-parsing/commit/d895babd90c3739ed41de53bb45ef873d56b1c37#commitcomment-61274147). I agree with @tgrapperon there that it isn't obvious that `Many(CharacterSet.alphanumerics).parse(" ")` won't terminate.

Since that comment, we aded a runtime check that should trigger a breakpoint and print to the console. Did you run into a case that doesn't hit that check? I hit a breakpoint when I run the example you gave.

> @stephencelis let us know your thoughts and I can close this PR if it's not the correct change. 🙂

I think this is probably the intended behavior for now, though we'd be interested in more discussion if you'd like to start a GitHub discussion!

> I also think there still should be some `CharacterSet` tests in the repo that can help document the right behaviors, so alternatively I can remove the nil change and keep the tests.

We'd definitely take any PRs that improve test coverage! We'll try to carve out some time for tests soon, though!

andrewjl:
> Since that comment, we aded a runtime check that should trigger a breakpoint and print to the console. Did you run into a case that doesn't hit that check? I hit a breakpoint when I run the example you gave.

Yes, I see the runtime check as well for that example too. Originally I was doing something different, using a `while let` instead of `Many`:

```swift
// This will never terminate unless it breaks on output == ""
while let output = characterSet.parse(&testInput) {
    print("Found something: \(output)")
} 
```

> I think this is probably the intended behavior for now, though we'd be interested in more discussion if you'd like to start a GitHub discussion!

Sounds good. I'd like to become more familiar with the library first. If I come up with something I will open a new PR for that discussion.

And in the meantime, I updated this PR to remove the behavior change and only include tests. The test for nil was replaced with one that validates that an empty string is returned instead, which is existing behavior. 

Happy to add more cases if anyone can think of some good ones.

stephencelis (approved):
Thanks @andrewjl!

PR #101, opened 2022-01-24
Misc fixes

stephencelis:
A few random items for post-release.


.spi.yml
```diff
@@ -1,5 +1,11 @@
 version: 1
 builder:
   configs:
+  - platform: ios
+    scheme: Parsing
+  configs:
+  - platform: tvos
+    scheme: Parsing
```
stephencelis, line 8: We had to do the same in [swift-identified-collections](https://github.com/pointfreeco/swift-identified-collections/commit/680bf440178a78a627b1c2c64c0855f6523ad5b9#diff-5ebda4968fa267b5715182dbe59912825960a0c52655551a4f1c022b712ceaae) to get the platforms to pick things up. I think otherwise it tries to build the command line tools.

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.4
```
stephencelis, line 1: We only support 5.4 now, so may as well be on the latest package tools.

Package.swift
```diff
@@ -22,14 +22,14 @@ let package = Package(
       name: "ParsingTests",
       dependencies: ["Parsing"]
     ),
-    .target(
+    .executableTarget(
```
stephencelis, line 25: Package tools 5.4 has a new type target type for executables.

Sources/swift-parsing-benchmark/StringAbstractions.swift
```diff
@@ -16,37 +16,35 @@ import Parsing
  */
 
 let stringAbstractionsSuite = BenchmarkSuite(name: "String Abstractions") { suite in
-  #if compiler(>=5.5)
```
stephencelis, line 19: Bringing back string abstractions with the fix you found for 5.4.

Sources/variadics-generator/VariadicsGenerator.swift
```diff
@@ -79,6 +79,7 @@ func outputForEach<C: Collection>(
   }
 }
 
+@main
```
stephencelis, line 82: Another thing that comes with 5.4.

mbrandonw (approved):

PR #107, opened 2022-02-08
Added reversion to original if any parsers in `Zip` variations throw an error

randomeizer:
Relates to Discussion #106.

### DIFF
--- a/Sources/Parsing/Builders/Variadics.swift
+++ b/Sources/Parsing/Builders/Variadics.swift
@@ -18,11 +18,13 @@ extension Parsers {
       P0.Output,
       P1.Output
     ) {
+      let original = input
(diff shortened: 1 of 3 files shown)


stephencelis (approved):
@randomeizer Thanks for taking the time to PR! We're actually in the middle of changing the backtracking behavior on `main`, and those changes will be taken into this branch soon. We're basically removing backtracking from all parsers with the exception of `Peek` and `Not`, and adding backtracking to each step of `OneOf` instead. This should make it much simpler to create parsers since you don't have to actively worry about backtracking work. It'll all be taken care of for you.

I'm going to merge this but just wanted to let you know it will be changing even more soon :smile:

PR #109, opened 2022-02-09
Use a `ResultBuilder` of `String` in VariadicGenerator

tgrapperon:
It is very possible that more intricate codegen will be needed in the future, and I feel this `@StringBuilder` that builds `String`'s lines by lines is already helping making things a little clearer.

For example:
```swift
output("extension OneOfBuilder {\n")
output("  @inlinable public static func buildBlock<")
outputForEach(0..<arity, separator: ", ") { "P\($0)" }
output(">(\n    ")
outputForEach(0..<arity, separator: ", ") { "_ p\($0): P\($0)" }
output("\n  ) -> Parsers.\(typeName)<")
outputForEach(0..<arity, separator: ", ") { "P\($0)" }
output("> {\n")
output("    Parsers.\(typeName)(")
outputForEach(0..<arity, separator: ", ") { "p\($0)" }
output(")\n  }\n}\n\n")
```
becomes:
```swift
"extension OneOfBuilder {"
"  @inlinable public static func buildBlock<\(list(0..<arity){ "P\($0)" })>("
"    " + list(0..<arity) { "_ p\($0): P\($0)" }
"  ) -> Parsers.\(typeName)<\(list(0..<arity){ "P\($0)" })> {"
"    Parsers.\(typeName)(\(list(0..<arity){ "p\($0)" }))"
"  }"
"}"
```

I'm using `for xx in` when no separator is needed, but it can feel a little weird when intertwined with the `list` function that works like `outputForEach ` (but with `", "` as default separator). Maybe some specific `forEach` function can be introduced to replace some `for in` loops with a syntax that is closer to `list`. Please let me know if there's some interest and I'll add one to the PR.

The code generated is the same as the current version of `Variadics.swift`, modulo some line wraps from `swift-format` with the latest `OneOfBuilder`'s.

### DIFF
--- a/Sources/variadics-generator/StringBuilder.swift
+++ b/Sources/variadics-generator/StringBuilder.swift
@@ -1,0 +1,39 @@
+@resultBuilder
+enum StringBuilder {
+  typealias Expression = String
+  typealias Component = [String]
+
+  static func buildExpression(_ expression: Expression) -> Component {
+    [expression]
+  }
+  static func buildBlock(_ components: Component...) -> Component {
+    components.flatMap { $0 }
+  }
+  static func buildArray(_ components: [Component]) -> Component {
+    components.flatMap { $0 }
+  }
+  static func buildOptional(_ component: Component?) -> Component {
+    component ?? []
+  }
+  static func buildEither(first component: Component) -> Component {
+    component
+  }
+  static func buildEither(second component: Component) -> Component {
+    component
+  }
+  static func buildLimitedAvailability(_ component: Component) -> Component {
+    component
+  }
+  static func buildFinalResult(_ component: Component) -> String {
+    component.joined(separator: "\n")
+  }
+}
+
+extension String {
+  // Using String.init is a little too much for the compiler.
+  static func build(
+    @StringBuilder lines: () -> String
+  ) -> String {
+    lines()
+  }
+}
--- a/Sources/variadics-generator/StringBuilder.swift
+++ b/Sources/variadics-generator/StringBuilder.swift
@@ -30,7 +30,7 @@ enum StringBuilder {
 }
 
 extension String {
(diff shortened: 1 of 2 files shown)


stephencelis:
Hi @tgrapperon! Thanks for taking the time to look into this. I agree it definitely reads better. Just a few things:

First, we have a lot of branches we've been synchronizing, all of which have churn in the variadics generator. I'm kinda of wary of merging any fundamental to the variadics generator till these branches have landed, which may be another month or two as we prepare things for release.

Second, to your point here:

> It is very possible that more intricate codegen will be needed in the future

We really hope that the opposite is true! Not sure if you've followed Apple's [swift-experimental-string-processing](https://github.com/apple/swift-experimental-string-processing) work, but they've been actively adding (experimental) support to Swift that would allow us to pretty much fully eliminate our dependency on code generation. Particularly, there is a new result builders feature called `buildBlock(combining:into:)` in the Swift compiler that is currently hidden by a feature flag. Hopefully they'll put it through the evolution process soon!

Because of this, we're thinking we may not want to invest too much time in maintaining the code generator, and any changes we do take we'd like to hold off on till the dust settles a bit. Wanna close this for now and we can revisit in a month or two?

tgrapperon:
No problem. I was aware of the progresses in `string-processing`, but it was mainly to fill the immediate need until then (or variadic generics), as the library seems to evolve faster than these other projects.
I've seen the churn of branches, and since I've seen some with unresolved conflicts in this specific file, I gave it a shot! The builder is straightforward, and I think it may have helped with many plans with the library (throwing, printing). It is also possible that more codegen will be needed with the hypothetical `Separated` parser (not necessary a new builder though). It also reads nicely in diffs, and merges are simpler.
You can always copy/paste it in some branch to check if it helps.
Thanks for the feedback!

PR #115, opened 2022-02-10
Remove extra `configs:`

finestructure:
See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1541

Because of the three `configs:` sections we're only getting the `watchos` section when reading the `.spi.yml` file. That's why iOS and tvOS are failing :)


stephencelis (approved):
😅 YAML

PR #119, opened 2022-02-11
Parser builder map transform throws

randomeizer:
Not proposing this gets forked exactly, more to illuminate a possible pathway. Open to better suggestions.


Sources/Parsing/Parsers/Map.swift
```diff
@@ -25,18 +25,18 @@ extension Parsers {
...
       self.upstream = upstream
       self.transform = transform
     }
 
     @inlinable
     @inline(__always)
-    public func parse(_ input: inout Upstream.Input) rethrows -> NewOutput {
-      self.transform(try self.upstream.parse(&input))
+    public func parse(_ input: inout Upstream.Input) throws -> NewOutput {
+      try self.transform(self.upstream.parse(&input))
```
stephencelis, line 39: Losing `rethrows` is probably the main reason why we'd want a dedicated operator for a throwing `map`. Combine does the same thing with `tryMap`. We're definitely open to adding `tryMap`! Want to start a discussion about `tryMap` and other potential throwing operators?

randomeizer: I've started one at #121.

PR #123, opened 2022-02-12
Add "separated" variants to `Parse`

tgrapperon:
Hello!

This is a return of #80 with a completely different implementation, but with the same rationale.
This approach is avoiding creating new types or overloads where unnecessary.

`ParserBuilder` is turned into a class, and a generic subclass `SeparatedParserBuilder<Separator>` is used to host many `buildFinalResult` overloads. They simply map each "inner" `ZipXXX` component `P` to a `ZipOV<P,Skip<Separator>>` and rewrap them in the same `ZipXXX` "OV" signature. In other words, it each component, but the last one, is turned into  (component+separator).

I initially created a dedicated `Separated` parser, but it occurred to me it would feel much more natural as `Parse` overloads. Furthermore, the underlying parsers are the same. The proposed syntax is thus:
```swift 
Parse {           //    Parse {
  Int.parser()    //      Int.parser()
  Double.parser() //     ";"
  "Hello!"        // ==   Double.parser()
} separator {     //     ";"
  ";"             //      "Hello!"
}                 //    }
```
I've tried to keep things as separated (!) as possible. A dedicated `Separated` parser can be implemented if `Parse` overloads should be avoided. `ParserBuilder` now being a class is unfortunately less elegant than an `enum`, but it is not instantiable by clients anyway. A completely dedicated `resultBuilder` type can be also used if this is causing issues somewhere.

As I'm understanding it, no change is required for the migration to throwing parsers.

Please let me know what you think!

### DIFF
--- a/Sources/Parsing/Builders/ParserBuilder.swift
+++ b/Sources/Parsing/Builders/ParserBuilder.swift
@@ -3,7 +3,9 @@
 ///
 /// See ``Parse`` for an entry point into this builder.
 @resultBuilder
-public enum ParserBuilder {
+public class ParserBuilder {
+  internal init() {}
+  
   @inlinable
   public static func buildBlock<P>(_ parser: P) -> P where P: Parser {
     parser
@@ -43,3 +45,13 @@ public enum ParserBuilder {
     .init(wrapped: parser)
   }
 }
+
+/// A specialization of ``ParserBuilder``, adapted to build parsers separated by a common separator.
+///
+/// See ``ParserBuilder`` for more information about this type, and the ``Parse`` parser `separated`
+/// overloads for an entry point into this builder.
+@resultBuilder
+public final class SeparatedParserBuilder<Separator>: ParserBuilder where Separator: Parser {
+  internal override init() { super.init() }
+}
+
--- a/Sources/Parsing/Builders/ParserBuilder.swift
+++ b/Sources/Parsing/Builders/ParserBuilder.swift
@@ -4,8 +4,6 @@
 /// See ``Parse`` for an entry point into this builder.
 @resultBuilder
 public class ParserBuilder {
-  internal init() {}
-  
   @inlinable
   public static func buildBlock<P>(_ parser: P) -> P where P: Parser {
     parser
@@ -51,7 +49,5 @@ public class ParserBuilder {
 /// See ``ParserBuilder`` for more information about this type, and the ``Parse`` parser `separated`
 /// overloads for an entry point into this builder.
 @resultBuilder
-public final class SeparatedParserBuilder<Separator>: ParserBuilder where Separator: Parser {
-  internal override init() { super.init() }
-}
+public final class SeparatedParserBuilder<Separator>: ParserBuilder where Separator: Parser {}
 
--- a/Sources/Parsing/Builders/Variadics.swift
+++ b/Sources/Parsing/Builders/Variadics.swift
@@ -6950,7 +6950,9 @@ extension Parsers {
   {
     public let p0: P0, p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6, p7: P7
 
-    @inlinable public init(_ p0: P0, _ p1: P1, _ p2: P2, _ p3: P3, _ p4: P4, _ p5: P5, _ p6: P6, _ p7: P7) {
+    @inlinable public init(
+      _ p0: P0, _ p1: P1, _ p2: P2, _ p3: P3, _ p4: P4, _ p5: P5, _ p6: P6, _ p7: P7
+    ) {
       self.p0 = p0
       self.p1 = p1
       self.p2 = p2
@@ -7046,7 +7048,9 @@ extension Parsers {
   {
     public let p0: P0, p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6, p7: P7, p8: P8
 
-    @inlinable public init(_ p0: P0, _ p1: P1, _ p2: P2, _ p3: P3, _ p4: P4, _ p5: P5, _ p6: P6, _ p7: P7, _ p8: P8) {
+    @inlinable public init(
+      _ p0: P0, _ p1: P1, _ p2: P2, _ p3: P3, _ p4: P4, _ p5: P5, _ p6: P6, _ p7: P7, _ p8: P8
+    ) {
       self.p0 = p0
       self.p1 = p1
       self.p2 = p2
@@ -7151,7 +7155,10 @@ extension Parsers {
   {
     public let p0: P0, p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6, p7: P7, p8: P8, p9: P9
 
-    @inlinable public init(_ p0: P0, _ p1: P1, _ p2: P2, _ p3: P3, _ p4: P4, _ p5: P5, _ p6: P6, _ p7: P7, _ p8: P8, _ p9: P9) {
+    @inlinable public init(
+      _ p0: P0, _ p1: P1, _ p2: P2, _ p3: P3, _ p4: P4, _ p5: P5, _ p6: P6, _ p7: P7, _ p8: P8,
+      _ p9: P9
+    ) {
       self.p0 = p0
       self.p1 = p1
       self.p2 = p2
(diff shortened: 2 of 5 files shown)


iampatbrown:
@tgrapperon how did you find this performed with regards to type checking? I've found using a custom parser/builder like you've mentioned more compiler friendly. It might also give a little more control over the error messaging. 

I was tempted to overload `Parse` with a `separator`, `initiator` and `terminator` or (`delimiter` when `initiator == terminator`) but ended up opting for the `Separated`/`Delimited` approach.

tgrapperon:
@iampatbrown Thanks for the feedback. I've experienced no issue with type checking. `SeparatedParserBuilder` is merely `ParserBuilder` with an extra `buildFinalResult` layer, but this layer is a 1 to 1 mapping from the `Zip`s to their "separated" counterparts, so there's no additional ambiguity and its straightforward once you require a to build a `(Separator) -> Parsers` function.
We can't reuse `ParserBuilder` as it directly, as we need the `Separator` generic from `SeparatedParserBuilder to build the final result (although I didn't check if it works as well with a non-generic builder type and an additional `Separator` generic at the builder function level). Implementing a builder from the ground up is also a solution, but I'm not seen which problem it'd solve. We can bypass the `buildFinalResult` but we will essentially reimplement all `ParserBuilder` but with different outputs. Did you experienced some issues?

When I've played with the builder, error messages were quite clear. The only thing that can happen is a mismatch of `Input` (and the message is clear, stating that XXX should be equivalent to YYY), or because you add more than 6 parsers, where `@ParserBuilder` error message are used.
Fun fact: if you explicitly state that `Separator.Input == Parsers.Input` for the `init`, you get unexpressive error messages in case of input mismatch ("type of expression is ambiguous without more context"). If you let the builder infer that (you only specify `Separator: Parser`), you get the mismatched types as mentioned above. 

I'm aware of the potential "terminator" variant, and I guess this approach should be able to support them without problems. I simply thought they could be implemented at the same time as `Many` gets a `terminator` too, but I can add them to this PR if the maintainers wish so. But the implementation should be discussed: Either we reuse the same builder overloads with an additional `Terminator` generic argument, and we specialise them using `Always<Void>` when separators/terminators are unspecified (less performant at runtime), or we ship a whole new set of overloads. But that would add 128 more overloads on top of the 64 this PR already adds. I don't know which direction is preferable. Maybe we can perform some logic around the `Always<Void>` in the `builder`s function body, and keep the number of overloads constant and the performance optimal.

I switched from a `Separated` type to `Parse` overloads when I realized I was reimplenting exactly `Parse` inner bits. This is fundamentally the same parser. When `separator:` is empty, this is equivalent to separate components using a non-consuming `Always<Void>`. Same if it has a `terminator`. This also plays in favor of reusing the same builder.

tgrapperon:
> I used `-warn-long-expression-type-checking=10` to see what was taking the longest.

The only delay I can imagine when building separated variants compared to `Parse` would be caused by the additional `buildFinalResult` layer. If it's really problematic, as dedicated builder that directly produces separated zips in `buildBlock` can be used without changing the interface for the user. I'll try to set up something to measure the differences.

> I was referring more to customizing error message thrown by the parser[…]

I thought about it too, and from what I've seen, you should get error message as clear as you would get when using a manual `Parse` with explicit separators. Right now, I can think of 2 ways to improve them if needed:
- Introduce a `Separator<S>` parser, functionally equivalent to `Skip<S>` that is used to skip the separators, but with a dedicated error message that states its a separator.
- Introduce a `Label<P>` parser that decorates its parser's errors with some user provided label. Then, with an extension on `Parser` we could have
``` swift
Double.parser().label("y-Coordinate")
```
The problem is that it'll add one level of indirection when parsing. I think it can be restricted to `#Debug` builds without affecting too much the type system in most cases, as the signature will essentially be the same. Or two methods, `label(_:)` and `debugLabel(_:)` could be introduced instead.
A better solution would be do define `ParserModifier` like in `SwiftUI`, but I haven't managed to solve the type system puzzle yet. The objective is to have `ParserModifier`s which are not generic over the modified parser. Many parsers in the library could be converted to modifiers.

> Would variadic generics influence you decision here?

The more I think about it, the more I think `Parse` is the "canonical" place to introduce separators (and terminators). I think that variadic generics may have made me miss that if they were already available, as I would have probably defined a `Separated` parser right away. If you look at the `print()` function from Swift, on a pure "call-site" point of view (even without making the parallel with parsers that Brandon and Stephen have mentioned). This is a method that we use very naturally to print one thing. We also use it a little less frequently to print two or more things. And even less frequently with a separator/terminator. At no point, the "separator" and "terminator" arguments are on the way when using simple `print(x)` or `print(x,y)`. Some users are even unaware of the existence of such arguments. It is also very clear to users that parsers used in `Parse` block are parsed in sequence, and a separator/terminator is then a very natural thing to expect. 
And the new trailing closure syntax is also helping in this case, as we simply append more arguments without touching the core "Parse {…}". 
The question was about the builder, and all this argumentation explains that's why I think the `SeparatedParserBuilder` should be a `ParserBuilder` itself.

An alternative would be to define `ZipN<P1, P2, Separator, Terminator>` and use specializations for the separator/terminator-less variants. I can experiment with this. We would end up with `ParserBuilder` being a typealias of `SeparatedTerminatedParserBuilder<EmptySeparator, EmptyTerminator>`. I think the total number of overloads will be the same, but I don't how much it'll affect build time. `Zip`s should probably be nested in the generic builder type itself and then `Separator, Terminator> arguments would go away, so it can maybe be OK:
```swift
@resultBuilder enum SeparatedTerminatedParserBuilder<Separator, Terminator> {
  struct ZipOV<P1, P2>: Parser {} …
}
```
I'll experiment with it. Especially if we can recover `ParserBuilder` as a typealias that works without generics. But this is mostly an implementation detail, unrelated to this PR.

tgrapperon:
> Using a class based builder is pretty interesting. I hadn't seen that before.
It also works with protocol extensions. You can't make the protocol be a builder, but you can host the overloads in extensions, and have the final enum being a builder and conforming to it:
```swift
protocol ParserBuilding {}
extension ParserBuilding {
  static func buildBlock(…)
}

@resultBuilder
enum ParserBuilder: ParserBuilding {}

@resultBuilder
enum SeparatedBuilder<Separator>: ParserBuilding where Separator: Parser {
  static func buildFinalResult(…)
}
```
I can refactor the PR to use something like this if needed.
It also means that we can compose builder behaviors using protocol composition. I don't know for this specific case, but this can maybe prove useful someday!

tgrapperon:
To sum up the current state of the PR: Right now, `Parse` gets new `separated` initializers, `ParserBuilder` is turned into a class and `SeparatedParserBuilder` is a subclass of it. The open questions are:
- Should a dedicated type be used instead of adding new initializers to `Parse`?
- Should a `terminated` variant be implemented too[^1]?
- Should we introduce a core `ParserBuilding` protocol and use its extensions to host combinatoric overloads instead of subclassing? `ParserBuilder` will go back being an `enum`, albeit conforming to `ParserBuilding`. The class/protocol conversion is relatively easy to implement, so I guess both should be tried and their build time measured at the end.

[^1]: If a `terminated` variant is introduced, I thing using dedicated builders for different configurations will allow to contain the number of overloads per builder type: `SeparatedParserBuilder`, `SeparatedTerminatedParserBuilder`, `TerminatedParserBuilder` (I don't know if this last configuration makes a lot of sense practically, but it should probably be there for coherence). In each type, the number of overloads will be constant (2^6=64 right now).

tgrapperon:
Thanks for the kind words @stephencelis! I have no issues with my PR not landing. I'm doing them mostly for the art/fun of pushing boundaries, and I'm often afraid to bother you with my experiments. I'm making PRs in the spirit of Swift Evolution, as it is always better to have some code to back a discussion.

I would also like to thank to @iampatbrown for reviewing it and providing, as always, very valuable comments and discussion points. I found the protocol-based builders thing by reflecting on his remark about using classes for builders.

Aside protocols, one of the main key-point to extract from this PR is the possibility to partially parametrize builders using generics and/or building functions as final result. One can also view this, from far away, as a way to defer builders results until we get the value of another argument in the function the builder is used (potentially a generic, and potentially another builder, like here). I know there are discussions about stateful builders, but until then, this can probably help in some cases.

About App Clips, I was thinking that all these overloads would be stripped if unused. I should probably experiment with some real use cases.

Thanks for your feedback!

iampatbrown:
> we also hope you realize that each one is quite exciting for us to look at, and they often push us to explore things we wouldn't have explored otherwise!

100% this :) Thanks heaps @tgrapperon I always enjoy exploring your PR's and learn a lot form them! While they might not get merged in their entirety, they definitely contribute to the libraries and benefit the community. Keep it up!

PR #129, opened 2022-02-15
Add `Separated` parser with `SeparableParser`

tgrapperon:
A new approach to the `Separated` parser (see #80 and #123). I'll sum it up as key-points:
- A new `SeparableParser: Parser` protocol is introduced. It supports a parse method with separator arguments. No new associated types as `Input` and `Output` are not affected by the separators (this is the key I think).
- All `Zip` parsers are `SeparableParser`. Their `Parser` requirement is provided by an extension of `SeparableParser` with `nil` separators (`nil` are used to denote the absence of separators for performance reasons)
- No overloads were added to the parser builder.
- `SeparableParser`s are exploited in a `Separated` parser  that accepts `initiator`, `separator` and `terminator`. There are probably too many `init`s, but it's easy to remove them if needed.
- An internal `Delimiter` parser is used to wrap separators and provide very nice error messages:
```swift
"error: unexpected input
 --> input:1:6
1 | 1,2,3;4
  |      ^ separator: expected ",""
```
- I've not noticed any performance impact. Benchmark results are similar.
- The archived benchmark app binary goes from 5 096 376 bytes to 5 859 240 bytes. That's a 15% increase. This can probably be lowered a bit without initiators, but this is significative.

This solution can effectively double what's a `Zip` parser is capable of handling under certain circumstances. A `Separated` parser with 6 components, an initiator, separator and terminator is parsing 13 parsers inline in one `Zip`.

A small issue persists right now: when writing the `Separated` parser, there is, at the beginning, one parser only. This parser is most likely not a `SeparableParser` that the type expects, so it complains and fails to build. 
I've now removed all `Separated` initializers without a `separator` parameter, so I guess it's natural to fail to build in this case. Otherwise, I don't know the best way to handle this. Maybe some disfavored unavailable init to provide a better message?

I'll write some documentation and more tests if this should go farther. I keep it as a draft for now. The 15% increase of size is concerning.

### DIFF
--- a/Sources/Parsing/Builders/Variadics.swift
+++ b/Sources/Parsing/Builders/Variadics.swift
@@ -1,10 +1,16 @@
 // BEGIN AUTO-GENERATED CONTENT
 
 extension Parsers {
-  public struct ZipOO<P0: Parser, P1: Parser>: Parser
+  public struct ZipOO<P0: Parser, P1: Parser>: SeparableParser
   where
     P0.Input == P1.Input
   {
+    public typealias Input = P0.Input
+    public typealias Output = (
+      P0.Output,
+      P1.Output
+    )
+
     public let p0: P0, p1: P1
 
     @inlinable public init(_ p0: P0, _ p1: P1) {
@@ -12,13 +18,23 @@ extension Parsers {
       self.p1 = p1
     }
 
-    @inlinable public func parse(_ input: inout P0.Input) rethrows -> (
+    @inlinable public func parse<Initiator, Separator, Terminator>(
+      _ input: inout P0.Input,
+      initiator: Initiator?, separator: Separator?, terminator: Terminator?
+    ) rethrows -> (
       P0.Output,
       P1.Output
-    ) {
+    )
+    where
+      Initiator: Parser, Separator: Parser, Terminator: Parser,
+      Initiator.Input == Input, Separator.Input == Input, Terminator.Input == Input
+    {
       do {
+        _ = try initiator?.parse(&input)
         let o0 = try p0.parse(&input)
+        _ = try separator?.parse(&input)
         let o1 = try p1.parse(&input)
+        _ = try terminator?.parse(&input)
         return (o0, o1)
       } catch { throw ParsingError.wrap(error, at: input) }
     }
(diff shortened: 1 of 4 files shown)


tgrapperon:
Out of curiosity, I've compared the benchmark binary size between #123 and the latest non throwing version. It goes from 4 968 456 to 5 217 000 bytes. A 5% increase (but without terminators/initiators)

stephencelis:
Very fun exploration to solve the problem! I think the same caveats apply for mergeability, but we look forward to reading through things nonetheless :smile:

tgrapperon:
I've tried several approaches to minimise the binary size, but inlining gets in the way each time. This solution generates bigger binaries because it has more logic than the `buildFinalResult` from #123. It is capable of handling more inner parsers in a simple `Zip` where `#123` would compose them to achieve the same throughput. This is a fortunate byproduct of the "overload-less" implementation, but #123 performance wasn't degraded a lot by the additional, 1-level `Zip` nesting, so it doesn't solve a real issue.

At the end of the day, inlining makes any general solution worse than a bespoke implementation for each parser. There is no hope in tunnelling every `Zip` body through a few common general implementations that are parsing tuples without filtering, and `Zip` picking what they want to output a result. Removing inlining would help making the binary smaller, but it would degrade even more the performance that is already impacted by the generalized implementation.

There is no real takeaway from this experiment. We generated bigger `Zip`s with a pre-baked structure to handle the common "delimited/separated" case, but it is roughly the same as having bigger general `Zip`s that are only used with a subset of arguments configurations. The idea from @iampatbrown of using a custom parser type to provide better error messages was very nice. I wonder if there is a way to implement this behavior without adding one level of indirection when parsing on the happy path.

So I'll close this for now. I think #123 approach is superior to this, regardless of the mergeability. Maybe I'll revisit the topic when printers are released.

Also, CI was failing on non-macos platforms, and it doesn't seem generalized to other PRs today. I'll check what can be the cause.

stephencelis:
@tgrapperon It's because #130 and #131 were merged at the same time, where one changed an error message tested in the other. I snuck a fix in in #113, which I was finally able to merge after wrestling with the Linux/Windows generic constraint solver.

PR #130, opened 2022-02-15
coalesce manyFailed error messages

mayoff:
In `ParsingError.debugDescription`, check for a `manyFailed` where all
the child errors are `ParsingError.failure` with the same
`originalInput` and `remainingInput` and no `underlyingError`s. In such
a scenario, only output the context once, for a more compact and legible
diagnostic.

Before:

    error: multiple failures occurred

    error: unexpected input
     --> input:1:1
    1 | president
      | ^ expected "admin"

    error: unexpected input
     --> input:1:1
    1 | president
      | ^ expected "guest"

    error: unexpected input
     --> input:1:1
    1 | president
      | ^ expected "member"

After:

    error: unexpected input
     --> input:1:1
    1 | president
      | ^ expected "admin"
      | ^ expected "guest"
      | ^ expected "member"

### DIFF
--- a/Sources/Parsing/ParsingError.swift
+++ b/Sources/Parsing/ParsingError.swift
@@ -152,28 +152,103 @@ extension ParsingError: CustomDebugStringConvertible {
   var debugDescription: String {
     switch self.flattened() {
     case let .failed(label, context):
-      return format(label: label, context: context)
+      return format(labels: [label], context: context)
 
     case let .manyFailed(errors, context) where errors.isEmpty:
-      return format(label: "", context: context)
+      return format(labels: [""], context: context)
 
     case let .manyFailed(errors, _):
-      let failures =
-        errors
-        .map(formatError)
-        .joined(separator: "\n\n")
+      return debugDescription(for: errors)
+    }
+  }
 
-      return """
+  fileprivate func debugDescription(for errors: [Error]) -> String {
+    func failed(_ error: Error) -> (String, Context)? {
+      guard
+        let error = error as? ParsingError,
+        case .failed(let label, let context) = error
+      else { return nil }
+      return (label, context)
+    }
+
+    var count = 0
+    var description = ""
+
+    func append(_ s: String) {
+      count += 1
+      if !description.isEmpty {
+        description.append("\n\n")
+      }
+      description.append(s)
+    }
+
+    var errors = errors[...]
+    while let error = errors.popFirst() {
+      guard case .some((let firstLabel, let firstContext)) = failed(error) else {
+        append(formatError(error))
+        continue
+      }
+
+      var labels = [firstLabel]
+      while
+        case .some((let label, let context)) = errors.first.flatMap({ failed($0) }),
+        firstContext.canGroup(with: context)
+      {
+        errors.removeFirst()
+        labels.append(label)
+      }
+
+      append(format(labels: labels, context: firstContext))
+    }
+
+    if count != 1 {
+      description = """
         error: multiple failures occurred
 
-        \(failures)
+        \(description)
         """
+    }
+
+    return description
+  }
+}
+
+extension ParsingError.Context {
+  fileprivate func canGroup(with other: Self) -> Bool {
+    func areSame(_ lhs: Any, _ rhs: Any) -> Bool {
+      switch (normalize(lhs), normalize(rhs)) {
+      case let (lhs as Substring, rhs as Substring):
+        return lhs.startIndex == rhs.startIndex && lhs.endIndex == rhs.endIndex
+      case let (lhs as Slice<[Substring]>, rhs as Slice<[Substring]>):
+        return zip(lhs, rhs).allSatisfy { l, r in
+          l.startIndex == r.startIndex && l.endIndex == r.endIndex
+        }
+      default:
+        return false
+      }
+    }
+
+    return (
+      areSame(self.remainingInput, other.remainingInput)
+      && areSame(self.originalInput, other.originalInput)
+      && self.underlyingError == nil && other.underlyingError == nil
+    )
+  }
+}
+
+extension Array where Element == (label: String, context: ParsingError.Context) {
+  fileprivate func allSame(_ keyPath: KeyPath<ParsingError.Context, Any>) -> Bool {
+    guard let first = self.first else { return true }
+    guard let firstValue = first.context[keyPath: keyPath] as? AnyHashable else { return false }
+    return self.dropFirst().allSatisfy {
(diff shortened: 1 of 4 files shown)


mayoff:
I expect there's more than just “a little” more work to do. 😉 I just wanted to get something working for a common scenario.

A `ParsingError.expected` case would be great, as would a `label` combinator.

The JSON example is interesting but questionable, for a couple of reasons:

- JSON doesn't allow multi-line strings. A more standard-complaint parser would fail at the newline on the `xs` line and therefore, I hope, end up with a single failure location.

- Even if JSON allowed multi-line strings, the errors at `input:5:9` are, IMO, spurious and confusing. It's as if I wrote `if a ~ b` and Swift not only complained about the `~`, but also suggested maybe I should have written `while` or `switch` instead of `if`.

Really, I have essentially the same complaint as @randomeizer makes [here](https://github.com/pointfreeco/swift-parsing/discussions/124#discussioncomment-2181480). `OneOf` goes overboard in combining errors. (So, I think, does `Many`.)
 
As soon as any alternative in a `OneOf` consumes input, the `OneOf` should commit to that alternative—even if that alternative fails. `OneOf` should only combine errors when **all** of the alternatives fail without consuming input. This strategy eliminates the spurious JSON errors.

I learned about this error-handling strategy from  [the Parsec paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/parsec-paper-letter.pdf). We'd also need a combinator (like Parsec's `try`) to hide that consumption in cases where deeper lookahead is required.

Is it even reasonable to fix `OneOf` the way I'd like, or is that even a direction you're interested in going? Have I gone completely off the rails here? (Spoiler alert: yes I have.)

stephencelis:
> I expect there's more than just “a little” more work to do. 😉 I just wanted to get something working for a common scenario.

I think you're closer than you think and that it's a matter of some kind of `group(while:)` over the `allSame` logic...though without digging in I may be underestimating 😅

> A `ParsingError.expected` case would be great, as would a `label` combinator.
> 
> The JSON example is interesting but questionable, for a couple of reasons:
> 
> * JSON doesn't allow multi-line strings. A more standard-complaint parser would fail at the newline on the `xs` line and therefore, I hope, end up with a single failure location.
> * Even if JSON allowed multi-line strings, the errors at `input:5:9` are, IMO, spurious and confusing. It's as if I wrote `if a ~ b` and Swift not only complained about the `~`, but also suggested maybe I should have written `while` or `switch` instead of `if`.

We had a `$0 != "\n"` check in the `Prefix`, which leads to a better primary error location, but a less-good error message (a more generic, less descriptive "failed to satisfy predicate"), so we kept things the way they are for now. In the future with some kind of `label` combinator we could address this, though!

> Really, I have essentially the same complaint as @randomeizer makes [here](https://github.com/pointfreeco/swift-parsing/discussions/124#discussioncomment-2181480). `OneOf` goes overboard in combining errors. (So, I think, does `Many`.)

Because we rank the errors, primarily based on how far along the branch consumes input, we could discard all failures beyond the primary "rank". We erred on the side of preserving _all_ failures for now, where hopefully the most important failure ranks highest and is the first one you see, and those below it are less and less relevant. We didn't want to accidentally discard failures that ranked lower but were more relevant.

> As soon as any alternative in a `OneOf` consumes input, the `OneOf` should commit to that alternative—even if that alternative fails. `OneOf` should only combine errors when **all** of the alternatives fail without consuming input. This strategy eliminates the spurious JSON errors.
>
> I learned about this error-handling strategy from [the Parsec paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/parsec-paper-letter.pdf). We'd also need a combinator (like Parsec's `try`) to hide that consumption in cases where deeper lookahead is required.
> 
> Is it even reasonable to fix `OneOf` the way I'd like, or is that even a direction you're interested in going? Have I gone completely off the rails here? (Spoiler alert: yes I have.)

It certainly sounds reasonable to explore! Errors are still very new to the library (and are internal for a reason). We've wanted them for a long time but held off for lack of prioritization. We expedited `throw`ing parsers and wanted a reasonable "alpha" for what we consider acceptable error messages, but we agree there is a lot of room for improvement and exploration.

mayoff:
ok fine… (grumble grumble)

Before:

```
error: multiple failures occurred

error: unexpected input
 --> input:6:4
6 |   "ys": {
  |    ^ expected ","

error: unexpected input
 --> input:6:4
6 |   "ys": {
  |    ^ expected "]"

error: unexpected input
 --> input:5:9
5 |   "xs": [1, "hello, null, false],
  |         ^ expected "{"

error: unexpected input
 --> input:5:9
5 |   "xs": [1, "hello, null, false],
  |         ^ expected "\""

error: unexpected input
 --> input:5:9
5 |   "xs": [1, "hello, null, false],
  |         ^ expected double

error: unexpected input
 --> input:5:9
5 |   "xs": [1, "hello, null, false],
  |         ^ expected "true" or "false"

error: unexpected input
 --> input:5:9
5 |   "xs": [1, "hello, null, false],
  |         ^ expected "null"

error: unexpected input
 --> input:4:19
4 |   "whatever": null,
  |                   ^ expected "}"

error: unexpected input
 --> input:1:1
1 | {
  | ^ expected "["

error: unexpected input
 --> input:1:1
1 | {
  | ^ expected "\""

error: unexpected input
 --> input:1:1
1 | {
  | ^ expected double

error: unexpected input
 --> input:1:1
1 | {
  | ^ expected "true" or "false"

error: unexpected input
 --> input:1:1
1 | {
  | ^ expected "null"
```

After:

```
error: multiple failures occurred

error: unexpected input
 --> input:6:4
6 |   "ys": {
  |    ^ expected ","
  |    ^ expected "]"

error: unexpected input
 --> input:5:9
5 |   "xs": [1, "hello, null, false],
  |         ^ expected "{"
  |         ^ expected "\""
  |         ^ expected double
  |         ^ expected "true" or "false"
  |         ^ expected "null"

error: unexpected input
 --> input:4:19
4 |   "whatever": null,
  |                   ^ expected "}"

error: unexpected input
 --> input:1:1
1 | {
  | ^ expected "["
  | ^ expected "\""
  | ^ expected double
  | ^ expected "true" or "false"
  | ^ expected "null"
```

stephencelis:
> ok fine… (grumble grumble)

😅 Just to be clear we absolutely are open to simplifying how errors accumulate in the future based on consumption! We just wanted to make sure we weren't missing anything or introducing situations where we were surfacing poor quality errors and totally ignoring higher-quality ones.

In the meantime, though, I think this is a great outward-facing change, and we can explore simplifying things before publicizing `ParsingError`.

PR #131, opened 2022-02-15
Prefix should eagerly consume

stephencelis:
We had a small bug in how the `Prefix` parser was consuming upon failure: it would always consume the minimum number of elements, which is more than the number of elements it actually consumed. This would produce bad error messages that point to the minimum successful position and _not_ the failure position.

### DIFF
--- a/Sources/Parsing/Parsers/Prefix.swift
+++ b/Sources/Parsing/Parsers/Prefix.swift
@@ -170,6 +170,7 @@ public struct Prefix<Input: Collection>: Parser where Input.SubSequence == Input
     var prefix = maxLength.map(input.prefix) ?? input
     prefix = predicate.map { prefix.prefix(while: $0) } ?? prefix
     let count = prefix.count
+    input.removeFirst(count)
     guard count >= self.minLength else {
       let atLeast = self.minLength - count
       throw ParsingError.expectedInput(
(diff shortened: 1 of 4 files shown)


Sources/Parsing/Parsers/Prefix.swift
```diff
@@ -170,17 +170,17 @@ public struct Prefix<Input: Collection>: Parser where Input.SubSequence == Input
...
     let count = prefix.count
+    input.removeFirst(count)
     guard count >= self.minLength else {
       let atLeast = self.minLength - count
       throw ParsingError.expectedInput(
         """
         \(self.minLength - count) \(count == 0 ? "" : "more ")element\(atLeast == 1 ? "" : "s")\
         \(predicate == nil ? "" : " satisfying predicate")
         """,
-        at: input.dropFirst(self.minLength)
```
stephencelis, line 180: This was dropping too many characters on failure.

Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -35,8 +35,10 @@ let jsonSuite = BenchmarkSuite(name: "JSON") { suite in
       string.append(contentsOf: fragment)
     } element: {
       OneOf {
-        Prefix(1...) { $0 != .init(ascii: "\"") && $0 != .init(ascii: "\\") }
-          .map { String(Substring($0)) }
+        Prefix(1...) {
+          $0 != .init(ascii: "\"") && $0 != .init(ascii: "\\") && $0 >= .init(ascii: " ")
```
stephencelis, line 39: This is "better" JSON parsing logic (control character like newlines are invalid). Before this change, the `Prefix(1...)` would push the failure onto the next line, which is more difficult to read/debug.

mbrandonw (approved):

PR #132, opened 2022-02-16
Use a closure `() -> Output` in `ReplaceError`

tgrapperon:
This PR adds a version for `replaceError` that takes a closure `(Error) -> Output`, so the user can adapt the result according to the type/value of the errors thrown. 
The constant version is preserved.

I'm preparing a sister PR (#135) that proposes a parser similar to `ReplaceError` but where the user can potentially rethrow another error, or fallback to an `output` value. This should allow the user to implement more logic on the failing side.

### DIFF
--- a/Sources/Parsing/Parsers/ReplaceError.swift
+++ b/Sources/Parsing/Parsers/ReplaceError.swift
@@ -33,8 +33,9 @@ extension Parser {
   /// - Parameter output: An output to return should the upstream parser fail.
   /// - Returns: A parser that never fails.
   @inlinable
-  public func replaceError(with output: Output) -> Parsers.ReplaceError<Self> {
-    .init(output: { _ in output }, upstream: self)
+  public func replaceError(with output: @autoclosure @escaping () -> Output)
+    -> Parsers.ReplaceError<Self> {
+    .init(output: { _ in output() }, upstream: self)
   }
   
   /// A parser that replaces its error with a provided output.
--- a/Sources/Parsing/Parsers/ReplaceError.swift
+++ b/Sources/Parsing/Parsers/ReplaceError.swift
@@ -35,18 +35,6 @@ extension Parser {
   @inlinable
   public func replaceError(with output: @autoclosure @escaping () -> Output)
     -> Parsers.ReplaceError<Self> {
-    .init(output: { _ in output() }, upstream: self)
-  }
-  
-  /// A parser that replaces its error with a provided output.
-  ///
-  /// See ``replaceError(with:)-1wzc1`` for more information.
-  ///
-  /// - Parameter output: A closure that returns an `Output` value when the parser fails and throws
-  /// the error provided as argument.
-  /// - Returns: A parser that never fails.
-  @inlinable
-  public func replaceError(with output: @escaping (Error) -> Output) -> Parsers.ReplaceError<Self> {
     .init(output: output, upstream: self)
   }
 }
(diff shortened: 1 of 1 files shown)


randomeizer:
It might be worth a) adding the `Input` into the closure, and b) having a couple of overloads of this function without the error/input parameters, in case we don't care and just want to output a default value regardless?

tgrapperon:
I just switched back the constant version an `@autoclosure`, as proposed in the discussions.

About `Input`, I don't know. It is bundled in the `ParsingError` (albeit internal right now). If the user throws their own errors, they can also probably host the `input`. 

Are there situations where you would exploit an input value coming along an opaque error to generate an output value?

randomeizer:
Well, ParsingError is currently not available outside the Parsing library. I don't have any current situations where I'm using it, but at one point had cloned the internal ParsingError in my own library, and it needed the input.

tgrapperon:
As I've closed #135, and without it, this PR makes much less sense until `ParsingError` is eventually made public.

tgrapperon:
Example implementation, following the benchmark results from https://github.com/pointfreeco/swift-parsing/discussions/134#discussioncomment-2194380.
I don't know if the box should be generic or embedded in `ReplaceError`. I went with the later.
I reopened only the PR so it can retrieve the new commit. I can close-it again without issues.
In any case, this is only an implementation detail for the user now.

tgrapperon:
Here are the benchmark results. "Uncached" is a variant from this PR implementation that only uses the function without storing the error on first throw and evaluate it each time (to measure the impact of setting/accessing the cache on the heap)

```
name                                           time          std        iterations
----------------------------------------------------------------------------------
ReplaceError.Current  - Create - Small Value       63.000 ns ± 254.08 %    1000000
ReplaceError.Proposed - Create - Small Value      148.000 ns ± 227.65 %    1000000
ReplaceError.Uncached - Create - Small Value       48.000 ns ± 566.95 %    1000000

ReplaceError.Current  - Create - Large Value   365011.000 ns ±  34.69 %       3465
ReplaceError.Proposed - Create - Large Value      147.000 ns ± 235.23 %    1000000
ReplaceError.Uncached - Create - Large Value       45.000 ns ± 346.49 %    1000000

ReplaceError.Current  - Run&Fail - Small Value    385.000 ns ± 108.63 %    1000000
ReplaceError.Proposed - Run&Fail - Small Value    440.000 ns ± 106.31 %    1000000
ReplaceError.Uncached - Run&Fail - Small Value    394.000 ns ± 120.43 %    1000000

ReplaceError.Current  - Run&Fail - Large Value    400.000 ns ± 233.11 %    1000000
ReplaceError.Proposed - Run&Fail - Large Value    483.000 ns ± 108.61 %    1000000
ReplaceError.Uncached - Run&Fail - Large Value 340533.000 ns ±  25.47 %       3738
```

To cache the value, I've tried to use a box (the current PR state), or rewriting the `output` function in the init using a local variable as storage, but boxing is faster in the init, and equivalent otherwise.

I can't draw any definitive conclusion. The impact of caching is still small and the performance is always bounded. The current implementation may spend a lot of time initialising (which is however not frequent usually), and the uncached function, at lot of time returning the result in case of failure (which can easily have dramatic effects).

I would lean toward closing again. Do you agree @stephencelis ?

tgrapperon:
Thinking again about the issue, the best outcome is reached when the user reuses a value they store somewhere else when it's something lengthly to create, and this is probably what they'll naturally do if they are handling such values.

So, I don't think it's worth the additional complexity and performance loss.

PR #139, opened 2022-02-18
Make ParsingError type publicly accessible

### LINKED ISSUE #138: Make `ParsingError` public?
I just started migrating over to the latest version of swift-parsing. While doing that, I had some custom parsers that I wanted to update to the new throwing style. One of my parsers has a `minCharacters` variable and when I can't find that chars, I'd like to throw a `ParsingError`. But when I do, I get an error because `ParsingError` seems to be internal to the library. Is there any reason we shouldn't have access to the `ParsingError` type for our custom parsers? While I could write my own error type, wouldn't it be more convenient if all parsers would throw the same errors?

Here's my custom parser:
```swift
/// A parser that consumes all whitespace characters from the beginning of the input. No output.
struct Whitespaces: Parser {
  let minCharacters: Int

  init(
    minCharacters: Int = 0
  ) {
    self.minCharacters = minCharacters
  }

  func parse(_ input: inout Substring) throws {
    let foundWhitespaces = input.prefix { char in
      char == " " || char == "\n" || char == "\r" || char == "\t"
    }

    guard foundWhitespaces.count >= minCharacters else {
      throw ParsingError.expectedInput("\(minCharacters) whitespaces", at: input)
      // => ERROR: Cannot find 'ParsingError' in scope
    }

    input.removeFirst(foundWhitespaces.count)
  }
}

```

Jeehut:
Fixes #138.

Don't know if this is a good implementation as I'm removing all those `@usableFromInline` statements, probably this makes the code less performant in some way. Should I have tried to make as many calls as possible `@inlineable` at least?

### DIFF
--- a/Sources/Parsing/ParsingError.swift
+++ b/Sources/Parsing/ParsingError.swift
@@ -1,12 +1,10 @@
 import Foundation
 
-@usableFromInline
-enum ParsingError: Error {
+public enum ParsingError: Error {
   case failed(String, Context)
   case manyFailed([Error], Context)
 
-  @usableFromInline
-  static func expectedInput(_ description: String, at remainingInput: Any) -> Self {
+  public static func expectedInput(_ description: String, at remainingInput: Any) -> Self {
     .failed(
       summary: "unexpected input",
       label: "expected \(description)",
@@ -14,8 +12,7 @@ enum ParsingError: Error {
     )
   }
 
-  @usableFromInline
-  static func expectedInput(
+  public static func expectedInput(
     _ description: String,
     from originalInput: Any,
     to remainingInput: Any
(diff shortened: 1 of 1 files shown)


tgrapperon:
I'm not an authority, but I don't think this PR will land right know. The authors prefer the `ParsingError` type to stay internal for now, to keep the freedom to make drastic changes to the type implementation/interface without having to handle backward compatibility or maintaining an ever-growing set of deprecations. It would probably become public at some point, once its shape has settled.

Functions that should execute quickly/frequently during parsing can be marked `@inlineable`. It means their implementation (and not only their existence) is known to clients of the library, allowing the compiler to copy the implementation instead of calling the function. For this reason, symbols that are touched from `@inlineable` functions should be accessible too. `public` fields are de facto usable from inline, as they're already part of the public API, and you use `@usableFromInline` it to make other internal symbols available from `@inlineable` functions. `@usableFromInline` doesn't provide any performance gain by itself. It's like a "semi-public" access rule (`@testable` is another attribute working with the same idea). `@inlineable` also implies `@usableFromInline`, but the converse is not true.

Once you mark it `@usableFromInline` or `public`, the symbol is part of the ABI, so this can be something to consider. This is usually not a concern until a 1.0 version is released.

PR #146, opened 2022-02-22
Fix build failure on Xcode 13.3b3 - Swift 5.6

tgrapperon:
`originalInput.base[…]` apparently now makes an `ArraySlice`, which is conflicting with the expected `Slice`. 
We can instantiate the `Slice` directly to avoid confusion.

This change should be backward-compatible.


stephencelis (approved):
Thanks for the fix!

PR #148, opened 2022-02-24
Optionally should backtrack when failing

randomeizer:
Like `OneOf`, `Optionally` should be backtracking on failure, since the wrapped `Parser` may consume some of the input before it fails.

I've also updated the `print` to throw any wrapped `print` failures, since if a value is provided, the wrapped printer shouldn't fail if it's valid.

### DIFF
--- a/Tests/ParsingTests/OptionallyTests.swift
+++ b/Tests/ParsingTests/OptionallyTests.swift
@@ -14,4 +14,13 @@ final class OptionalTests: XCTestCase {
     XCTAssertNoDifference(.none, Optionally { Bool.parser() }.parse(&input))
     XCTAssertNoDifference("Hello, world!", Substring(input))
   }
+    
+  func testBacktracking() {
+    var input = "Hello, world!"[...]
+    XCTAssertNoDifference(.none, Optionally {
+      "Hello, "
+      Bool.parser()
+    }.parse(&input))
+    XCTAssertNoDifference("Hello, world!", input)
+  }
 }
--- a/Tests/ParsingTests/OptionallyTests.swift
+++ b/Tests/ParsingTests/OptionallyTests.swift
@@ -16,11 +16,16 @@ final class OptionalTests: XCTestCase {
   }
     
   func testBacktracking() {
-    var input = "Hello, world!"[...]
-    XCTAssertNoDifference(.none, Optionally {
-      "Hello, "
-      Bool.parser()
-    }.parse(&input))
-    XCTAssertNoDifference("Hello, world!", input)
+    let parser = Parse {
+      "Hello,"
+      Optionally {
+        " "
+        Bool.parser()
+      }
+      " world!"
+    }
+    
+    XCTAssertNoDifference(.some(true), try parser.parse("Hello, true world!"))
+    XCTAssertNoDifference(.none, try parser.parse("Hello, world!"))
   }
 }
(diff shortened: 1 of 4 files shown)


mbrandonw:
Hi @randomeizer, I'm curious why you think `Optionally` should backtrack. With the changes in #108 we decided that `OneOf` should be the sole place for backtracking.

One argument I could see for backtracking in `Optionally` is because technically it could be rewritten as a `OneOf` mapping into `Optional`:

```swift
Optionally { p }

// vs

OneOf {
  p.map(Optional.some)
  Always(nil)  
}
```

But I'm curious if you came across a concrete example of where it would be expected.

Tests/ParsingTests/OptionallyTests.swift
```diff
@@ -14,4 +14,18 @@ final class OptionalTests: XCTestCase {
...
+      Optionally {
+        " "
+        Bool.parser()
+      }
+      " world!"
+    }
+    
+    XCTAssertNoDifference(.some(true), try parser.parse("Hello, true world!"))
+    XCTAssertNoDifference(.none, try parser.parse("Hello, world!"))
+  }
```
randomeizer, line 30: I've updated the `testBacktracking` test case to illustrate where `Optionally` will not parse as expected if there is no backtracking.

The first assertion of `"Hello, true world!"` will pass with or without backtracking, because the `Optionally` doesn't fail and correctly consumes the `" true"` string (including the leading space).

The second assertion of `"Hello, world!"` will fail without backtracking, because the space will be consumed inside the `Optionally` and will not be available for the `" world!"` parser.

mbrandonw: Nice, that does make a lot of sense. Only other option I can think of to consider is whether we introduce an explicit `.backtracking` operator to opt into backtracking, and continue leaving all other backtracking to just `OneOf`. But the fact that `Optionally` is just sugar for `OneOf` does make me think it should backtrack.

So, I do think this is a good change, but will wait to merge it in for a few days because @stephencelis is traveling right now. Also I think we should make this change on main and then bring into the printer branch separately since this doesn't depend on anything from printing. Could you rebase it when you have a chance?

randomeizer: Yeah, I think it produces more unexpected results if there is no backtracking. I was scratching my head for a bit initially when it popped up in my parser.

My changes do tweak the `print` method also, and throws an error if the print fails, although that is perhaps a different issue.

randomeizer: In general, I think that any `Parser` which catches errors and may continue on without throwing its own error should backtrack. Otherwise you will end up with surprising results for parsers further down the line.

mbrandonw: Yeah I think we can bring the print changes back once the parse stuff is in main. And then we can worry about tests and documentation for that behavior at that time too.

In fact, I suppose docs could be updated to mention the backtracking behavior, but I can take a stab at that once it merges if you don't have time.

mbrandonw: > In general, I think that any `Parser` which catches errors and may continue on without throwing its own error should backtrack. Otherwise you will end up with surprising results for parsers further down the line.

Yeah that's interesting. We should keep that in mind for future documentation. We're working on a [DocC](https://github.com/pointfreeco/swift-parsing/compare/docc) overhaul so that we have better places to call out info like this.

randomeizer:
I've created as separate PR at #149 against `main` which just updates the `parse` method to backtrack, adds the unit test, and updates the documentation. You'll probably want to review the text. I'll close this one down.

PR #149, opened 2022-02-25
Added backtracking to Optionally

randomeizer:
Also updated documentation about backtracking in general.

The `OptionallyTest.testBacktracking()` case gives an example of where Optionally will cause an unexpected parsing error if it doesn't backtrack. The second assert:

```swift
XCTAssertEqual(.none, try parser.parse("Hello, world!"))
```

will fail without it, since the `" "` before `"world"` will be consumed within the `Optionally`.


randomeizer:
Ok, great. And agreed, if a failure happens and is passed on, no backtracking is required.

mbrandonw:
I just merged `main` into this PR and once it goes green I'll merge! Thanks again @randomeizer for the help!

PR #158, opened 2022-03-01
Backtracking Finesse

stephencelis:
After discussion from #149, we have a better handle on when and where a parser should backtrack. Basically: if a parser can recover from failure, it should backtrack to the `input`'s state before the failure occurred.

This means we can finesse the behavior of backtracking in the few parsers that need to consider it.

- `Optionally` — Not implemented in this PR, but we will merge #149 in soon!
- `replaceError(with:)` — This already had backtracking baked into it because it's a specialization of `OneOf`, though this PR adds a test.
- `Not` — This parser (along with `Peek`) _always_ backtracked, since they shouldn't consume input. But we think there's no reason for `Not` to _always_ backtrack. If it fails it should be safe enough to leave input consumed and rely on upstream `OneOf`s, etc., to do the work of backtracking.
- `OneOfMany` — `OneOf` only backtracks when recovering input to run another parser, which means it will not perform the extra work of backtracking on the final parser should it fail. `OneOfMany`, however, was always backtracking, even when its final parser failed. This PR has added logic to avoid this extra work.

### DIFF
--- a/Tests/ParsingTests/NotTests.swift
+++ b/Tests/ParsingTests/NotTests.swift
@@ -36,7 +36,7 @@ class NotTests: XCTestCase {
     XCTAssertEqual(
       input,
       """
-      // a comment
+       a comment
       let foo = true
       """
     )
--- a/Sources/Parsing/Parsers/Not.swift
+++ b/Sources/Parsing/Parsers/Not.swift
@@ -26,10 +26,10 @@ public struct Not<Upstream: Parser>: Parser {
   @inlinable
   public func parse(_ input: inout Upstream.Input) throws {
     let original = input
-    defer { input = original }
     do {
       _ = try self.upstream.parse(&input)
     } catch {
+      input = original
       return
     }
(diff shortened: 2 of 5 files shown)


mbrandonw (approved):
Nice, should we get tests for `Not`'s and `OneOfMany`'s backtracking behavior?

stephencelis:
> Nice, should we get tests for `Not`'s and `OneOfMany`'s backtracking behavior?

Done!

Tests/ParsingTests/NotTests.swift
```diff
@@ -36,7 +36,7 @@ class NotTests: XCTestCase {
     XCTAssertEqual(
       input,
       """
-      // a comment
+       a comment
```
mbrandonw, line 39: did you mean to make this change?

stephencelis: Yeah this tests that `Not` is consuming.

PR #160, opened 2022-03-01
Remove duplicate docs from README and Parser.

mbrandonw:
The DocC articles are now the source-of-truth for parser info, and so we can remove a bunch of duplicated info from the README and `Parser` docs.


stephencelis (approved):
Looks good!

PR #161, opened 2022-03-01
Auto-trim trailing markdown whitespace

stephencelis:
Since Xcode doesn't trim markdown files, let's automate this at the CI level.


mbrandonw (approved):
Noice

PR #162, opened 2022-03-01
Adjusted Peek

randomeizer:
No longer returns the matched peek value.
Adjusted backtracking behaviour to match new expectations
Added backtracking test.


stephencelis (approved):
Looks great! Thanks again :smile:

PR #167, opened 2022-03-03
Updates Optionally to throw any internal `wrapped.print(...)` error

randomeizer:
Added unit tests for printing, including example of why it should fail if provided a bad "output" to print.


stephencelis (approved):
@randomeizer Good catch! Especially now that we think of `Optionally` as sugar around `OneOf`, we could produce the equivalent parser from your test case:

```swift
OneOf {
  Prefix { $0 != " " }.map(.case(Optional.some))
  Always(()).map { Substring?.none }
}
```

This parser will indeed fail to print `.some("foo bar")`, so `Optionally` should do the same! Thanks for digging into the problem :smile:

PR #168, opened 2022-03-03
DocC typo fix

KeithBird:
Learning this library through DocC is very enjoyable.

I read all the articles and found some typos. They are in markdown's code blocks so they are difficult to detect.

And I'm confused that a symbol (such as `Parse`) sometimes wrapped in a set of backticks (\`Parse\`) and sometimes wrapped in a set of double backticks (\`\`Parse\`\`). I recommend using double backticks whenever possible to help readers jump to definition and alleviate difficulty in choosing. 😝


stephencelis (approved):
Good finds, thanks!

PR #163, opened 2022-03-02
Adding more structure and hierarchy to the documentation catalogue

icanzilb:
It's exciting to see you use DocC to create cool documentation for this library. The current state of the documentation on "main" is mostly using the default curation grouping APIs by their type (enumeration, functions, etc.) — I tried to clean up the main page a bit and add grouping that would help developers discover the APIs more easily.

This is an umbrella PR of a bunch of changes done according to the guidelines in https://github.com/apple/swift-docc/tree/main/Sources/DocCDocumentation/DocCDocumentation.docc, https://github.com/apple/swift-docc/blob/main/Sources/DocCDocumentation/DocCDocumentation.docc/adding-structure-to-your-documentation-pages.md specifically but also the rest.

I haven't used `Parsing` extensively so I expect I mis-filed a bunch of APIs so I'm looking forward to your feedback and doing iterations over the current changes.

PR #171, opened 2022-03-03
DocC updates

mbrandonw:
A few pass at taking some of the ideas from @icanzlib's PR #163. More to come soon, but this gets some basics in which should help with our printer documentation.


Sources/Parsing/Builders/OneOfBuilder.swift
```diff
@@ -1,7 +1,21 @@
-/// A custom parameter attribute that constructs a parser that attempts to run a number of parsers,
-/// one after the other, till one succeeds with an output.
+/// A custom parameter attribute that constructs parsers from closures. The constructed parser
+/// runs each parser in the closure, one after another, till one succeeds with an input.
```
mbrandonw, line 2: Took some of this wording from the `@ViewBuilder` documentation.

The "custom parameter attribute" is weird to me. Kinda feel like we could say something similar...

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
```diff
@@ -0,0 +1,66 @@
...
+
+On the other hand, if `Bool.parser()` is used in a context where the input type cannot be inferred,
+then you will get an compiler error:
+
+```swift
+try Parse {
+  Bool.parser()
+  Bool.parser() // 🛑 Ambiguous use of 'parser(of:)'
+}
+.parse("truefalse")
```
mbrandonw, line 54: now we have a place to document some of this ambiguity behavior

Sources/Parsing/Documentation.docc/Extensions/Parser.md
```diff
@@ -0,0 +1,68 @@
...
+Some of the most commonly used parsers in the library. Use these parsers with operators in order
+to build complex parsers from simpler pieces.
+
+- <doc:Int>
+- <doc:String>
+- <doc:Bool>
+- <doc:Double>
+- <doc:Float>
+- <doc:CharacterSet>
+- <doc:UUID>
```
mbrandonw, line 32: now we have a place to document this stuff that isn't currently visible to DocC

Sources/Parsing/Builders/OneOfBuilder.swift
stephencelis: Do you mean say something other than "custom parameter attribute"?

stephencelis, line 2: I think previously this meant "succeed by returning an input"

```suggestion
/// runs each parser in the closure, one after another, till one succeeds with an output.
```

But I can see this also being "succeed to parse an input". So depending on which you want to focus on...

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
```diff
@@ -0,0 +1,66 @@
+# Bool
+
+A parser that consumes a Boolean value from the beginning of a string.
+ 
+This parser only recognizes the literal "true" and "false" characters:
```
stephencelis, line 5:
```suggestion
This parser only recognizes the literal "true" and "false" sequence of characters:
```

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
```diff
@@ -0,0 +1,66 @@
...
+
+// error: unexpected input
+//  --> input:1:1
+// 1 | 1 Hello
+//     ^ expected "true" or "false"
+```
+
+The `Bool.parser()` method is overloaded to work on a variety of string representations in order
+to be as efficient as possible, including `Substring`, `UTF8View`, and generally collections of 
+UTF-8 code units (see <doc:StringAbstractions> for more info).
```
stephencelis, line 30:
```suggestion
to be as efficient as possible, including `Substring`, `UTF8View`, and more general collections of 
UTF-8 code units (see <doc:StringAbstractions> for more info).
```

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
stephencelis: Does that snippet actually fail even with `.parse`? I assume that's enough info, so I guess it's the result builder bug?

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
```diff
@@ -0,0 +1,66 @@
...
+.parse("truefalse")
+```
+
+To fix this you can force one of the boolean parsers to be the `Substring` parser, and then the 
+other will figure it out via type inference:
+
+```swift
+try Parse {
+Bool.parser(of: Substring.self)
+Bool.parser()
```
stephencelis, line 63:
```suggestion
  Bool.parser(of: Substring.self)
  Bool.parser()
```

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
stephencelis: (Take this or leave it. I think the current way reads OK on seeing it in the other docs, but it did read strange to me initially.)

Sources/Parsing/Parser.swift
```diff
@@ -56,7 +65,63 @@ extension Parser {
...
+  /// ```
+  ///
+  /// > Tip: If your input can have trailing whitespace that you would like to consume and discard
+  /// > you can do so like this:
+  /// > ```swift
+  /// > let output = try Parse {
+  /// >  Int.parser()
+  /// >  ",".utf8
+  /// >  Bool.parser()
+  /// >  Skip { Whitespace() }
```
stephencelis, line 121:
```suggestion
  /// >   Int.parser()
  /// >   ",".utf8
  /// >   Bool.parser()
  /// >   Skip { Whitespace() }
```

Sources/Parsing/Parser.swift
```diff
@@ -69,7 +134,62 @@ extension Parser {
...
+  /// ```
+  ///
+  /// This method will fail if the parser does not consume the entirety of the input.
+  /// For example:
+  ///
+  /// ```swift
+  /// let output = try Parse {
+  ///  Int.parser()
+  ///  ",".utf8
+  ///  Bool.parser()
```
stephencelis, line 172:
```suggestion
  ///   Int.parser()
  ///   ",".utf8
  ///   Bool.parser()
```

stephencelis: Should this be in a `> Tip:` like above?

stephencelis (approved):
Looking really good!

Sources/Parsing/Parser.swift
mbrandonw: I thought we were leaving this part as-is and just making the whitespace thing a tip. You think both should be tips?

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
mbrandonw: Oh no it does not, I'll break it out into two lines.

PR #173, opened 2022-03-04
Improvements to allow throwing initialisers to be mapped.

randomeizer:
So, I've been building out an XML Parser, and I'm at the point where I want to start validating input. The simplest way to achieve that is to have throwing initialisers, and currently the library doesn't really allow that via the normal `.map(...)` or `Parse(Foo.ini)` methods currently.

This PR has an implementation of `tryMap`, as well as making the `.struct`/`Conversions.Struct` methods `throws`-friendly.

However, although I know we discussed it previously, I'm going to advocate against `tryMap` as a separate implementation and for making `Parser.map(_:)` take a `throws` closure. Here are my reasons:

1. While it's "clean" to be able to use `.map(_:)` to create a non-throwing `Parser`, in practice any slightly complicated composite parser will throw regardless. It doesn't have the same value as a non-throwing `.map()` in, say, Combine.
2. The fix for making structs 'throwing' was to modify the `.struct`/`Conversions.Struct` conversion section. I looked at Creating a `.tryStruct` and `TryStruct` to handle the throwing version, noticed that the 'Struct.unapply()` throws anyway. Also, no working code that actually executes `Struct.apply(...)` does so directly, so it has to `try` anyway.

To sum up, while I have a `.tryMap` implemented here, I would rather that the standard `.map` can take a throwing closure. If you think it still has value in some cases, I guess an `.alwaysMap()` could be added?

Do you have any examples of where mapping to a non-throwing Parser is valuable? I haven't been able to find one.


randomeizer:
Just to add to this, I wasn't planning on pushing the throwing `.map()` version, but when I did, it literally only required one line of code to be changed outside the `Map.swift` implementation:

```swift
XCTAssertNoDifference(43, try Always(42).map { $0 + 1 }.parse(&input))
```

Just had to add the `try` here.

stephencelis:
It's definitely a bummer that there's no way for operators like `Parser.map` to be rethrowing, and I hope whenever rethrowing protocols are formally pitched, this shortcoming is addressed.

> [...] any slightly complicated composite parser will throw regardless. It doesn't have the same value as a non-throwing `.map()` in, say, Combine.
>
> Do you have any examples of where mapping to a non-throwing Parser is valuable?

Some of the more complicated language parsers out there, like programming language parsers, should never throw, and this is done via a liberal use of error recovery (for example, via `replaceError(with:)`).

A few links about error resilience/recovery and parsing:

  * [Syntax error recovery in parsing expression grammars](https://dl.acm.org/doi/10.1145/3167132.3167261): a paper referenced in [this nom blog post](https://eyalkalderon.com/blog/nom-error-recovery/) about error recovery.
  * [Constructing Human Grade Parsers](https://duriansoftware.com/joe/constructing-human-grade-parsers): a post by Joe Groff about how parsers should always return a partial successful result, even on failures
  * [chumsky](https://github.com/zesterer/chumsky): a Rust parsing library built with error resilience as a primary goal

In adopting `rethrows`, swift-parsing is in a great position to make it easy to know when your parser is resilient or not. For this reason, we'd love it if operations that are non-failing, like `map`, don't suddenly promote non-throwing parsers into throwing ones. If anything, we might explore how to make existing throwing parsers into non-throwing ones, but we'd like to avoid doing the opposite.

Really we do think parsers and publishers are quite similar, and so we think that `tryMap` vs. `map` has the same value here as it has in Combine.

In addition, `try`/`throws` does incur a performance cost, so providing non-throwing versions of common operations can make a difference here. (Though, to be honest, Swift's performance characteristics do not appear to always be easy to predict.)

> To sum up, while I have a `.tryMap` implemented here, I would rather that the standard `.map` can take a throwing closure. If you think it still has value in some cases, I guess an `.alwaysMap()` could be added?

I think we'd like to lean on `map`'s properties here, and the prior art of Combine's `map` vs. `tryMap` distinction.

We'll take a look at the body of this PR soon, though! We definitely think the operation is a useful addition!

Sources/Parsing/ParserPrinters/Map.swift
```diff
@@ -15,7 +15,7 @@ extension Parser {
   @_disfavoredOverload
   @inlinable
   public func map<NewOutput>(
-    _ transform: @escaping (Output) -> NewOutput
+    _ transform: @escaping (Output) throws -> NewOutput
```
stephencelis, line 18: I think we want to revert the changes in this file.

Sources/Parsing/ParserPrinters/TryMap.swift
```diff
@@ -0,0 +1,90 @@
...
+  ///
+  /// Mapping a parser with a transform function results in a parser that cannot print. If you need
+  /// to transform the output of a parser and retain printability, use the ``Conversion``-based
+  /// ``map(_:)-2sblf`` operator instead.
+  ///
+  /// - Parameter transform: A closure that transforms values of this parser's output.
+  /// - Returns: A parser of transformed outputs.
+  ///
+  ///
+  @_disfavoredOverload
```
stephencelis, line 19:
```suggestion
```

This is only necessary on operations that are also present on `Sequence`, since `String` is both a parser and a sequence.

randomeizer:
I'll hopefully have a chance to read some of your links soon.

Question: What are your thoughts on making `.struct()`/`Struct` throwing, as in my changes here?

stephencelis:
> Question: What are your thoughts on making `.struct()`/`Struct` throwing, as in my changes here?

I think that change is fine! Unfortunately, `@rethrows` is not smart enough to know when it needs to `throw` (`unapply`) vs. when it doesn't (`apply`), so `.struct` already incurs the throwing cost.

iampatbrown:
> we'd love it if operations that are non-failing, like map, don't suddenly promote non-throwing parsers into throwing ones.

@stephencelis can this be resolved by overloading `.map` instead of introducing `.tryMap`? In the same way `.map` may result in `Parsers.MapConstant<Self, NewOutput>` or `Parsers.MapConversion<Self, C>` there could also be:

```swift
@_disfavoredOverload
@inlinable
public func map<NewOutput>(
  _ transform: @escaping (Output) throws -> NewOutput
) -> Parsers.TryMap<Self, NewOutput> {
  .init(upstream: self, transform: transform)
}
```

> we think that tryMap vs. map has the same value here as it has in Combine. 

I don't mind tryMap. I thought it mainly existed for explicitly and to avoid adding additional overloads to map. Are there additional benefits I'm overlooking? 

> Swift's performance characteristics do not appear to always be easy to predict.

I haven't been able to reliably benchmark Map vs TryMap. I've tried a few different implementations that have been inconclusive.

Sources/Parsing/ParserPrinters/TryMap.swift
randomeizer: Ah ok. I'm not familiar with the annotation, so just left it in place after copy-pasting from `Map`.

randomeizer:
Ok, thanks for the links, super helpful. Spent a fair bit of time reading through them (other than the research paper, which seems to be under lock and key - looks interesting though). They have convinced me that `map` should definitely throw.

Having parsers that never fail is great. In fact, before the throwing version of the parser library came out, I was going down the path of capturing error results within the output tree, rather than returning `nil`, so that error messages wouldn't be lost. After reading the links above, I'll probably resume that approach.

However, `map` should not be the tool to achieve that. Performing a `map` is essentially a hand-rolled parser. It takes input values, does some transformations, and returns a new result. It should be allowed to fail, otherwise it will be forced to do error handling in a hand-rolled manner also.

For example, let's look at the one line of code I had to adjust when adding `throws` to the map, in its original form:

```swift
XCTAssertNoDifference(43, Always(42).map { $0 + 1 }.parse(&input))
```

This works because we know we're starting with `42` and we can add `1` to it without an error. However, if I changed it to this:

```swift
XCTAssertNoDifference(Int.max, Always(Int.max).map { $0 + 1 }.parse(&input))
```

...it crashes due to an overflow. There is a bug in the `map`. It actually needs to do some error checking. Ok, let's make it work:

```swift
XCTAssertNoDifference(Int.max /* + 1??? */, Always(Int.max).map {
  guard $0 < Int.max else {
    // um...
  }
  return $0 + 1 
}.parse(&input))
```

I guess one way might be to just return an optional value:

```swift
XCTAssertNoDifference(.none, Always(Int.max).map { input -> Int? in
  guard input < Int.max else {
    return .none
  }
  return input + 1
}
```

Ok, now it actually compiles and runs, but not ideal in most cases, since now we've lost the reason why it's `nil`. Ok, lets return a `Result` instead:

```swift
let addOne = UInt8.parser().map { input -> Result<Int,MyError> in
  guard input < Int.max else {
    return .failure(MyError.intOverflow)
  }
  return .success(input + 1)
}
```

Ok, cool, now I'm at least getting either a `.success` or `.failure`. However:

1. It's much less composable. I've now got to unwrap the successful result, do something with it and also handle the error case separately.
2. In a larger parsing operation, we now have no context on where the actual bad number occurred. The `map` function has no access to the `input`, so can't include it in the `Result`.3. 

---

**TL;DR:** `map` is a parser, and should be expected to throw errors sometimes. As such, `map` calls should throw, and `tryMap` is not necessary.

---

All that said, being able to create a "never fails" parser would be great. It would be great if we had something along the lines of the `expect` function from [eyal kalderon's](https://eyalkalderon.com/blog/nom-error-recovery/) example, which would let you do something like:

```swift
let addOne = Int.parser().map {
  guard $0 < Int.max
  else { throw MyError.intOverload }
  return $0 + 1
}.expect("value to be less than \(Int.max)")

var input = "\(Int.max)"[...]
XCTAssertNoDifference(ParseResult(.nil, ParsingError.manyFailed(...), addOne.parse(&input))
```

Even better is if we could mix multiple "expecting" parsers, and have the results/errors regrouped:

```swift
struct Email {
  let username: Substring
  let server: Substring
}

let username = Parse {
  // check for valid characters, etc.
}

let serverAddress = Parse {
  // check for valid characters, etc.
}

let email = Parse(Email.init) {
  username
  "@".expect("@")
  Prefix(1...).expect("server address")
}

var input = "foo@bar"[...]
let result = email.parse(&input)
```

And then this would always parse, no try required, and the "expectations" would be logged as errors.

But that is beyond the scope of this PR, I believe.

stephencelis:
> Ok, thanks for the links, super helpful. Spent a fair bit of time reading through them (other than the research paper, which seems to be under lock and key - looks interesting though). They have convinced me that `map` should definitely throw.
> 
> Having parsers that never fail is great. In fact, before the throwing version of the parser library came out, I was going down the path of capturing error results within the output tree, rather than returning `nil`, so that error messages wouldn't be lost. After reading the links above, I'll probably resume that approach.
> 
> However, `map` should not be the tool to achieve that. Performing a `map` is essentially a hand-rolled parser. It takes input values, does some transformations, and returns a new result. It should be allowed to fail, otherwise it will be forced to do error handling in a hand-rolled manner also.

If `rethrows` were preserved I think we'd mostly agree, but as it stands, we think the vast majority of `map`s should be simple transform functions (like bundling values up into a `struct` or `enum`, or performing some other non-failing transform), so having `map` always turn non-failing parsers into failing ones would make the simplest of transforms costly and introduce a question of error recovery into cases where there should be none.

> For example [...]
> 
> ```swift
> XCTAssertNoDifference(Int.max, Always(Int.max).map { $0 + 1 }.parse(&input))
> ```
> 
> ...it crashes due to an overflow. There is a bug in the `map`. It actually needs to do some error checking. Ok, let's make it work:

This example would make any `map` crash, no? An array, optional, result, publisher, etc., would all crash here if you fail to account for overflow.

When it comes to arrays and optionals and results, the transform is immediate and `rethrows` works, so it makes sense for ergonomics' sake to allow those closures to rethrow. But when it comes to publishers, a `map` that can throw would mean transforming publishers that never fail into a failing ones, which is just not what we want for the vast majority of simple transforms.

Because of this, it makes sense (for now) to have a dedicated operation for `map` and for `tryMap` (for publishers _and_ parsers). In the future, if `rethrows` can ever be relied on for such cases, maybe we could squash the two operations into one.

> I guess one way might be to just return an optional value

In this case you'd probably use `compactMap`.

> The best option (in my opinion) is to throw the error

And here you could change things to a `tryMap`.

This isn't to say that other parser operators couldn't be upgraded to take throwing closures if those closures are primarily there to encode failure. For example, `filter` should probably be throwing, as should `compactMap`.

randomeizer:
I've spent some time reading

> > Ok, thanks for the links, super helpful. Spent a fair bit of time reading through them (other than the research paper, which seems to be under lock and key - looks interesting though). They have convinced me that `map` should definitely throw.
> If `rethrows` were preserved I think we'd mostly agree, but as it stands, we think the vast majority of `map`s should be simple transform functions (like bundling values up into a `struct` or `enum`, or performing some other non-failing transform), so having `map` always turn non-failing parsers into failing ones would make the simplest of transforms costly and introduce a question of error recovery into cases where there should be none.

Can you give me an example of a simple, non-failing parser that uses `map` in a real-world example? There are none in your codebase bar the example below, and there are none in mine, because they are all fed by throwing parsers further up the chain.

> > For example [...]
> > ```swift
> > XCTAssertNoDifference(Int.max, Always(Int.max).map { $0 + 1 }.parse(&input))
> > ```
> >   
> > ...it crashes due to an overflow. There is a bug in the `map`. It actually needs to do some error checking. Ok, let's make it work:
> 
> This example would make any `map` crash, no? An array, optional, result, publisher, etc., would all crash here if you fail to account for overflow.

My point original point was, the only example in the codebase that is a non-thrown parser using `map` actually should be throwing an error when it encounters an overflow. It then seemed to make a good example for exploring error handling within a map, and building a non-throwing parser, so I stuck with it :)
 
> When it comes to arrays and optionals and results, the transform is immediate and `rethrows` works, so it makes sense for ergonomics' sake to allow those closures to rethrow. But when it comes to publishers, a `map` that can throw would mean transforming publishers that never fail into a failing ones, which is just not what we want for the vast majority of simple transforms.

Again, can you provide a real-world example of this?

> Because of this, it makes sense (for now) to have a dedicated operation for `map` and for `tryMap` (for publishers _and_ parsers). In the future, if `rethrows` can ever be relied on for such cases, maybe we could squash the two operations into one.
> 
> > I guess one way might be to just return an optional value
> 
> In this case you'd probably use `compactMap`.

I guess? I don't think returning `nil` in this example was a good choice though.

> > The best option (in my opinion) is to throw the error
> 
> And here you could change things to a `tryMap`.
> 
> This isn't to say that other parser operators couldn't be upgraded to take throwing closures if those closures are primarily there to encode failure. For example, `filter` should probably be throwing, as should `compactMap`.

And if not those, why not `map`? The `Converter` version of `map` already throws. From a pure API POV, does it not seem weird going from:

```swift
Parse {
  // parse stuff...
}.tryMap(MyStruct.init)
```

to:

```swift
ParsePrint {
  // parse stuff...
}.map(.struct(MyStruct.init))
```

Just because you upgraded it to a `Printer`?

randomeizer:
Hey there again, hope you had a great weekend. I've pushed an update where I revert `map` to non-throwing, basically for the purpose of benchmarking, and added some benchmarks for `Map`.

Here's a typical result.

```
name                               time        std        iterations
--------------------------------------------------------------------
Map.Always map                       35.000 ns ± 763.98 %    1000000
Map.Always tryMap                    43.000 ns ± 570.20 %    1000000
Map.Always tryMap replaceError       45.000 ns ± 657.07 %    1000000
Map.Int map                          99.000 ns ± 331.00 %    1000000
Map.Int tryMap                       99.000 ns ± 495.65 %    1000000
Map.Int map replaceError            100.000 ns ± 387.26 %    1000000
Map.Int tryMap replaceError         107.000 ns ± 344.24 %    1000000
Map.Int tryMap throws              3305.000 ns ±  64.17 %     401387
Map.Int tryMap throws replaceError 3392.000 ns ±  71.33 %     397839
```

You can check out the benchmarks for the details, but basically:

- `Always map`: the only parser that is completely non-throwing
- `Always tryMap`: the throwing `tryMap` forces the parser to be throwing
- `Always tryMap replaceError`: adding `replaceError` makes the final parser non-throwing, but it's still throwing internally.
- `Int map`: input is via an `Int.parser()`, then mapped via non-throwing `map`
- `Int tryMap`: input is via a `Int.parser(), then mapped via throwing `tryMap`
- `Int map replaceError`: `Int.parser()` mapped via non-throwing `map` then `replaceError`
- `Int tryMap replaceError`: `Int.parser()` mapped via throwing `tryMap` then `replaceError`
- `Int tryMap throws`: `Int.parser()` succeeds then `tryMap` just throws an error
- `Int tryMap throws replaceError`: `Int.parser()` succeeds then `tryMap` throws and error which is replaced by `replaceError`.

The takeaway:

1. The only real difference is between `Always map` and `Always tryMap`, where `tryMap` has about a 16-17% penalty.
2. Once the non-throwing `map` is wrapped in a throwing parser (eg. `Int.parser()`), there is no real difference between it and `tryMap`. It seems that any `try` will add a penalty, but it seems like it's a one-off, and internal try/catch statements don't stack the penalty.
3. Actually throwing errors is super expensive. Are we sure this is the best way to send error signals? `OneOf` is potentially super-expensive, for example.

iampatbrown:
> Actually throwing errors is super expensive.

This is most likely due to `ParsingError` still being internal. My guess is that making it public or optimising the current implementation would improve your benchmarks.

randomeizer:
> > Actually throwing errors is super expensive.
> 
> This is most likely due to `ParsingError` still being internal. My guess is that making it public or optimising the current implementation would improve your benchmarks.

Hmm, perhaps. Just created a `OneOf` benchmark.

Results with a private `ParsingError`:

```
name        time        std        iterations
---------------------------------------------
OneOf.one    256.000 ns ± 211.99 %    1000000
OneOf.two    644.000 ns ± 151.81 %    1000000
OneOf.three 1279.000 ns ±  81.73 %    1000000
OneOf.fails 1977.000 ns ±  54.86 %     673772
```

Results with a public `ParsingError`:

```
name        time        std        iterations
---------------------------------------------
OneOf.one    274.000 ns ± 329.43 %    1000000
OneOf.two    634.000 ns ±  88.08 %    1000000
OneOf.three 1266.000 ns ± 100.31 %    1000000
OneOf.fails 2006.000 ns ±  83.63 %     575267
```

Not much difference there. Just pushed my benchmark code, comments welcome.

iampatbrown:
> Results with a public ParsingError

Pretty sure the internal functions are probably preventing it from being inlined. Or did you make those public as well? I can have a look a bit later if you'd like. On phone atm. Making it all public or just replacing the Error inside `TryMap` with something basic might be one way to confirm.

randomeizer:
I only made the error (and related `Context`) public. Which other functions would need to be made public? Just the ones on `ParsingError`? Not exactly sure why, since nothing outside uses them directly?

The above results were from the `OneOf` benchmark. I'll run it against the `Map` benchmark also though.

randomeizer:
Made everything that was `@usableFromInline` public, and got basically the same results for the `Map` benchmark as well:

```
name                               time        std        iterations
--------------------------------------------------------------------
Map.Always map                       35.000 ns ± 792.87 %    1000000
Map.Always map memberwise            43.000 ns ± 600.02 %    1000000
Map.Always tryMap                    42.000 ns ± 477.98 %    1000000
Map.Always tryMap replaceError       46.000 ns ± 602.93 %    1000000
Map.Int map                          97.000 ns ± 259.30 %    1000000
Map.Int map memberwise               96.000 ns ± 298.06 %    1000000
Map.Int tryMap                       97.000 ns ± 528.24 %    1000000
Map.Int map replaceError             99.000 ns ± 497.94 %    1000000
Map.Int tryMap replaceError         103.000 ns ± 376.90 %    1000000
Map.Int tryMap throws              3284.000 ns ±  61.89 %     412928
Map.Int tryMap throws replaceError 3303.000 ns ±  50.97 %     404131
```

randomeizer:
Ok, so found the major contributor in the `Map` benchmark: My `TryMap` was wrapping the thrown error in a `ParsingError`, which was blowing out the numbers. When just letting the error go up the chain it's still quite expensive, but not so bad:

```
name                               time       std        iterations
-------------------------------------------------------------------
Map.Always map                      38.000 ns ± 658.26 %    1000000
Map.Always map memberwise           48.000 ns ± 307.35 %    1000000
Map.Always tryMap                   33.000 ns ± 467.19 %    1000000
Map.Always tryMap replaceError      45.000 ns ± 575.90 %    1000000
Map.Int map                         97.000 ns ± 466.24 %    1000000
Map.Int map memberwise              98.000 ns ± 353.39 %    1000000
Map.Int tryMap                      95.000 ns ± 436.50 %    1000000
Map.Int map replaceError           104.000 ns ± 621.06 %    1000000
Map.Int tryMap replaceError        104.000 ns ± 298.70 %    1000000
Map.Int tryMap throws              168.000 ns ± 364.51 %    1000000
Map.Int tryMap throws replaceError 200.000 ns ± 321.69 %    1000000
```

This is more in line with the `OneOf` benchmark. Throwing seems to cost an additional 100ns (on my computer).

The reason I put that in was because if not, the context of the error is getting lost when emerging out the other side directly from a `tryMap`. So I kind of feel like that's correct behaviour. I suspect there are opportunities for optimising `ParsingError.wrap()` as well though...

randomeizer:
So, no `tryMap` or anything? I think we need some implementation of it, even if `map` is non-throwing.

randomeizer:
Understood. Should I open another PR against the main branch?

PR #174, opened 2022-03-04
Fix some docc and added docs to Not and Peek.

mbrandonw:
A docc fix that @KeithBird pointed out, and also I noticed that we didn't have any actual usage of the `Not` and `Peek` parsers so I added some more to their docs.


KeithBird:
On the basis of this commit I have created a new PR #175.
To avoid conflicts, we can merge my PR directly maybe.

PR #175, opened 2022-03-04
Fix recently updated docc files & added docs to Not and Peek

KeithBird:
Previous this kind of code can be compiled (while the original intention is not so), because Xcode can infer that the type is `Substring` by `.parse(".1.2")`.

```swift
try Parse {
  Double.parser()
  Double.parser() // 🛑 Ambiguous use of 'parser(of:)'
}
.parse(".1.2")
```

We can fix it like this:

```swift
let parser = Parse {
  Double.parser()
  Double.parser() // 🛑 Ambiguous use of 'parser(of:)'
}

try parser.parse(".1.2")
```

### DIFF
--- a/Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
@@ -2,7 +2,7 @@
 
 A parser that consumes a Boolean value from the beginning of a string.
 
-This parser only recognizes the literal "true" and "false" sequence of characters:
+This parser only recognizes the literal `"true"` and `"false"` sequence of characters:
 
 ```swift
 // Parses "true":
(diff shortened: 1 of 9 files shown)


Sources/Parsing/Documentation.docc/Articles/Parsers/Int.md
```diff
@@ -42,36 +42,39 @@ UTF-8 code units (see <doc:StringAbstractions> for more info).
...
 other will figure it out via type inference:
 
 ```swift
-try Parse {
-  Int.parser(of: Substring.self)
-  Bool.parser()
+let parser = Parse {
+  Int.parser // ✅
+  Bool.parser(of: Substring.self)
 }
```
KeithBird, line 77: When `Int.parser()` and `Bool.parser()` are called in one block, type inference will fail.

Compiler Error:
`Type '(Substring.UTF8View.Type, Bool, Int) -> Parsers.IntParser<Substring.UTF8View, Int>' cannot conform to 'Parser'`

Sources/Parsing/Documentation.docc/Articles/Parsers/Bool.md
```diff
@@ -31,16 +31,17 @@ UTF-8 code units (see <doc:StringAbstractions> for more info).
...
 ```swift
-try Parse {
+let parser = Parse {
   Bool.parser()
   ","
   Bool.parser()
 }
-.parse("true,false") // (true, false)
+
+try parser.parse("true,false") // (true, false)
```
mbrandonw, line 44: Thanks! Stephen brought this to my attention before merging #171 but looks like I didn't update all the spots.

Sources/Parsing/Documentation.docc/Articles/Parsers/CharacterSet.md
```diff
@@ -5,7 +5,7 @@ A parser that consumes the characters contained in a `CharacterSet` from the beg
 For example:
 
 ```swift
-Parse {
+try Parse {
```
mbrandonw, line 8: This can actually be called without `try` because `CharacterSet` is a [non-throwing](https://github.com/pointfreeco/swift-parsing/blob/f22452a0e2e96d195a6ef5bfa1b8f91c0aa02470/Sources/Parsing/Parsers/CharacterSet.swift#L5) parser.

```suggestion
Parse {
```

mbrandonw (approved):
Great catches! Thanks!

PR #176, opened 2022-03-04
Add parser for case-iterable, string raw-representable values

stephencelis:
Backporting and updating this from another branch.

### DIFF
--- a/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
@@ -26,7 +26,7 @@ try Parse {
 This also works with raw enumerations that are backed by integers:
 
 ```swift
-enum Role: String, CaseIterable {
+enum Role: Int, CaseIterable {
   case admin = 1
   case guest = 2
   case member = 3
--- a/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
@@ -1,0 +1,73 @@
+# CaseIterable
+
+A parser that consumes a case-iterable, raw representable value from the beginning of a string.
+
+Given a type that conforms to `CaseIterable` and `RawRepresentable` with a `RawValue` of `String`
+or `Int`, we can incrementally parse a value of it.
+
+Notably, raw enumerations that conform to `CaseIterable` meet this criteria, so cases of the
+following type can be parsed with no extra work:
+
+```swift
+enum Role: String, CaseIterable {
+  case admin
+  case guest
+  case member
+}
+
+try Parse {
+  Int.parser()
+  ","
+  Role.parser()
+}
+.parse("123,member") // (123, .member)
+```
+
+This also works with raw enumerations that are backed by integers:
+
+```swift
+enum Role: String, CaseIterable {
+  case admin = 1
+  case guest = 2
+  case member = 3
+}
+
+try Parse {
+  Int.parser()
+  ","
+  Role.parser()
+}
+.parse("123,1") // (123, .admin)
+```
+
+The `parser()` method on `CaseIterable` is overloaded to work on a variety of string representations 
+in order to be as efficient as possible, including `Substring`, `UTF8View`, and more general
+collections of UTF-8 code units (see <doc:StringAbstractions> for more info).
+
+Typically Swift can choose the correct overload by using type inference based on what other parsers
+you are combining `parser()` with. For example, if you use `Role.parser()` with a
+`Substring` parser, like the literal "," parser in the above examples, Swift
+will choose the overload that works on substrings.
+
+On the other hand, if `Role.parser()` is used in a context where the input type cannot be inferred,
+then you will get an compiler error:
+
+```swift
+let parser = Parse {
+  Int.parser()
+  Role.parser() // 🛑 Ambiguous use of 'parser(of:)'
+}
+
+try parser.parse("123member")
+```
+
+To fix this you can force one of the boolean parsers to be the `Substring` parser, and then the
+other will figure it out via type inference:
+
+```swift
+try Parse {
+  Int.parser(of: Substring.self)
+  Role.parser()
+}
+.parse("123member") // (123, .member)
+```
--- a/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
@@ -61,7 +61,7 @@ let parser = Parse {
 try parser.parse("123member")
 ```
 
-To fix this you can force one of the boolean parsers to be the `Substring` parser, and then the
+To fix this you can force one of the parsers to be the `Substring` parser, and then the
 other will figure it out via type inference:
 
 ```swift
--- a/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
@@ -65,9 +65,9 @@ To fix this you can force one of the parsers to be the `Substring` parser, and t
 other will figure it out via type inference:
 
 ```swift
-try Parse {
+let parser = Parse {
   Int.parser(of: Substring.self)
   Role.parser()
 }
-.parse("123member") // (123, .member)
-```
+
+try parser.parse("123member") // (123, .member)
--- a/Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
+++ b/Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
@@ -100,10 +100,7 @@ extension Parsers {
     Prefix.Element == Input.Element
   {
     @usableFromInline
-    let cases: [Output]
-
-    @usableFromInline
-    let toPrefix: (Output.RawValue) -> Prefix
+    let cases: [(case: Output, prefix: Prefix, count: Int)]
 
     @usableFromInline
     let areEquivalent: (Input.Element, Input.Element) -> Bool
@@ -113,17 +110,20 @@ extension Parsers {
       toPrefix: @escaping (Output.RawValue) -> Prefix,
       areEquivalent: @escaping (Input.Element, Input.Element) -> Bool
     ) {
-      self.toPrefix = toPrefix
       self.areEquivalent = areEquivalent
-      self.cases = Output.allCases.sorted(by: { $0.rawValue > $1.rawValue })
+      self.cases = Output.allCases
+        .map {
+          let prefix = toPrefix($0.rawValue)
+          return ($0, prefix, prefix.count)
+        }
+        .sorted(by: { $0.count > $1.count })
     }
 
     @inlinable
     public func parse(_ input: inout Input) throws -> Output {
-      for `case` in self.cases {
-        let prefix = self.toPrefix(`case`.rawValue)
+      for (`case`, prefix, count) in self.cases {
         if input.starts(with: prefix, by: self.areEquivalent) {
-          input.removeFirst(prefix.count)
+          input.removeFirst(count)
           return `case`
         }
       }
--- a/Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
+++ b/Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
@@ -1,4 +1,13 @@
 extension CaseIterable where Self: RawRepresentable, RawValue == Int {
+  /// A parser that consumes a case-iterable, raw representable value from the beginning of a
+  /// collection of a substring.
+  ///
+  /// See <doc:CaseIterable> for more info.
+  ///
+  /// - Parameter inputType: The `Substring` type. This parameter is included to mirror the
+  ///   interface that parses any collection of UTF-8 code units.
+  /// - Returns: A parser that consumes a case-iterable, raw representable value from the beginning
+  ///   of a substring.
   @inlinable
   public static func parser(
     of inputType: Substring.Type = Substring.self
@@ -6,6 +15,15 @@ extension CaseIterable where Self: RawRepresentable, RawValue == Int {
     .init(toPrefix: { String($0) }, areEquivalent: ==)
   }
 
+  /// A parser that consumes a case-iterable, raw representable value from the beginning of a
+  /// collection of a substring's UTF-8 view.
+  ///
+  /// See <doc:CaseIterable> for more info.
+  ///
+  /// - Parameter inputType: The `Substring.UTF8View` type. This parameter is included to mirror the
+  ///   interface that parses any collection of UTF-8 code units.
+  /// - Returns: A parser that consumes a case-iterable, raw representable value from the beginning
+  ///   of a substring's UTF-8 view.
   @inlinable
   public static func parser(
     of inputType: Substring.UTF8View.Type = Substring.UTF8View.self
@@ -13,6 +31,12 @@ extension CaseIterable where Self: RawRepresentable, RawValue == Int {
     .init(toPrefix: { String($0).utf8 }, areEquivalent: ==)
   }
 
+  /// A parser that consumes a case-iterable, raw representable value from the beginning of a
+  /// collection of UTF-8 code units.
+  ///
+  /// - Parameter inputType: The collection type of UTF-8 code units to parse.
+  /// - Returns: A parser that consumes a case-iterable, raw representable value from the beginning
+  ///   of a collection of UTF-8 code units.
   @inlinable
   public static func parser<Input>(
     of inputType: Input.Type = Input.self
@@ -29,23 +53,7 @@ extension CaseIterable where Self: RawRepresentable, RawValue == String {
   /// A parser that consumes a case-iterable, raw representable value from the beginning of a
   /// collection of a substring.
   ///
-  /// Given a type that conforms to `RawRepresentable` with a `RawValue` of `String`, and
-  /// `CaseIterable`, we can incrementally parse a value of it.
-  ///
-  /// Notably, raw enumerations that conform to `CaseIterable` meet this criteria, so cases of the
-  /// following type can be parsed with no extra work:
-  ///
-  /// ```swift
-  /// enum Role: String, CaseIterable {
-  ///   case admin
-  ///   case guest
-  ///   case member
-  /// }
-  ///
-  /// var input = "admin,123"[...]
-  /// try Role.parser().parse(&input)  // Role.admin
-  /// input                            // ",123"
-  /// ```
+  /// See <doc:CaseIterable> for more info.
   ///
   /// - Parameter inputType: The `Substring` type. This parameter is included to mirror the
   ///   interface that parses any collection of UTF-8 code units.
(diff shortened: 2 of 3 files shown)


Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
```diff
@@ -0,0 +1,106 @@
...
+    Input.SubSequence == Input,
+    Input.Element == UTF8.CodeUnit
+  {
+    .init(toPrefix: { $0.utf8 }, areEquivalent: ==)
+  }
+}
+
+extension Parsers {
+  public struct CaseIterableRawRepresentableParser<
+    Input: Collection, Output: CaseIterable & RawRepresentable, Prefix: Collection
```
stephencelis, line 68: I've been able to relax some constraints by adding a separate generic for the prefix type that is used to test the `starts(with:)` logic in `parse`. So the type is a lil more intense than in the printer branch...

mbrandonw:
I just added a [commit](https://github.com/pointfreeco/swift-parsing/pull/176/commits/69a1eb856edf2e8ab76efe7efb948364402e6482) that brings over the `CaseIterable.md` article from the printer branch, and I realized we could make this stuff work when the raw value is an integer too:

```swift
enum Person: Int, CaseIterable {
  case blob = 4
  case blobJr = 42
}

let peopleParser = Many {
  Person.parser()
} separator: {
  ",".utf8
} terminator: {
  End()
}

var input = "4,42"[...].utf8
XCTAssertEqual(try peopleParser.parse(&input), [.blob, .blobJr])
```

Sources/Parsing/Documentation.docc/Articles/Parsers/CaseIterable.md
```diff
@@ -0,0 +1,73 @@
...
+  ","
+  Role.parser()
+}
+.parse("123,member") // (123, .member)
+```
+
+This also works with raw enumerations that are backed by integers:
+
+```swift
+enum Role: String, CaseIterable {
```
tgrapperon, line 29:
```suggestion
enum Role: Int, CaseIterable {
```

mbrandonw: Oops thanks!

Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
```diff
@@ -0,0 +1,133 @@
+extension CaseIterable where Self: RawRepresentable, RawValue == Int {
```
iampatbrown, line 1: @mbrandonw would `RawValue: LosslessStringConvertible & Comparable` also work here?

tgrapperon: @iampatbrown If I'm not mistaken, the restriction can theoretically be narrowed to being expressible by an integer or floating-point literal, because that's ultimately what will be allowed for `RawRepresentable` enums (other than those with `RawValue` being `ExpressibleByStringLiteral)`. So we can cover every situation with 3 overloads (`ExpressibleBy(Integer/Float/String)Literal`), but I don't know if we can easily extract parser/printers from it.

Being `Comparable` allows to reach `UInt`s for example, but it doesn't help with custom made types. For example, this compiles:
```swift
struct Value: RawRepresentable, Equatable {
  var rawValue: Int
  init(rawValue: Int) { self.rawValue = rawValue }
}

extension Value: ExpressibleByIntegerLiteral {
  init(integerLiteral value: Int) {
    self.rawValue = value
  }
}

enum Enum: Value, CaseIterable {
  case first = 1
}
```

`String` and `Int` are already covering the vast majority of the cases (!) though, but I 'm afraid that `LosslessStringConvertible & Compable` will only partially improve the situation (the `Enum` above would pass through).

Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
```diff
@@ -0,0 +1,133 @@
...
+    let areEquivalent: (Input.Element, Input.Element) -> Bool
+
+    @usableFromInline
+    init(
+      toPrefix: @escaping (Output.RawValue) -> Prefix,
+      areEquivalent: @escaping (Input.Element, Input.Element) -> Bool
+    ) {
+      self.toPrefix = toPrefix
+      self.areEquivalent = areEquivalent
+      self.cases = Output.allCases.sorted(by: { $0.rawValue > $1.rawValue })
```
tgrapperon, line 118: Wouldn't it be sufficient to sort by length instead? This would allow dropping the `Comparable` requirement, isn't it?

Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
iampatbrown: @tgrapperon if I'm not mistaken, the reason `RawValue == Int` works is because `Int` conforms to `LosslessStringConvertible & Comparable`.  The `String.init` being used in `toPrefix: { String($0) }` is `init<T>(_ value: T) where T : LosslessStringConvertible`. 

`RawValue: LosslessStringConvertible & Comparable` should work for `FixedWidthInteger`, `Double` and `Substring`. As for printing, I'm pretty sure that would remain the same using `toPrefix`:

```swift
input.prepend(contentsOf: self.toPrefix(output.rawValue))
```

Maybe I'm missing something though :P

Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
iampatbrown: @tgrapperon as in something like this?

```swift
self.cases = Output.allCases.sorted(by: {
  toPrefix($0.rawValue).endIndex > toPrefix($1.rawValue).endIndex
})
```

Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
tgrapperon: @iampatbrown You're right about `LosslessStringConvertible`. I was only reasoning at the `RawRepresentable` enum level, without considering how we can extract parsers/printers from the `RawValue` later.

You're probably right for the constraint, and the example I've provided is quite exotic and a dedicated parser can always be implemented if it ever happens in the wild. Furthermore, the new `RawValue` types that can be reached are all comparable, so it directly works with respect to the comment I made about sorting below.

I didn't check if the compiler was working the same with `Int`/`String` or `LosslessStringConvertible&Comparable`, but if it is, it looks like a better alternative.

Sources/Parsing/Parsers/CaseIterableRawRepresentable.swift
iampatbrown: I guess it doesn't really matter if they have the same length. The main issue would be if a shorter value is ordered before a longer one. This probably means that comparing `.endIndex` would probably work for `Double` but comparing `.rawValue` might not

iampatbrown: @tgrapperon actually the current implementation would fail for negative `Int`s too. Maybe using `endIndex` is actually the way to go. You can change the current test to negative values to test it out. 

```swift
enum Person: Int, CaseIterable {
  case blob = -4
  case blobJr = -42
}

var input = "-4,-42"[...].utf8
XCTAssertEqual(try peopleParser.parse(&input), [.blob, .blobJr]) // XCTAssertEqual failed
```

tgrapperon: @iampatbrown With the current state of the PR, this is not really a concern. Sorting is useless but mostly harmless for numeric `RawValue`s, and the other types are collections where sorting by `count` would be equivalent to bin the `lexical` sort.
So it's working indeed, but it performs a little more work than needed. The `Comparable` requirement is fortuitously fulfilled by all supported `RawValue` types. Being `LosslessStringConvertible` automatically provides a canonical way to sort values by projecting them to `.description`, so it can maybe be possible to drop the `Comparable` requirement for a simple `LosslessStringConvertible`, and perform the sort via `.description` only when needed if we manage to exclude numbers (maybe by checking if they're `IntegerLiteralType` or `FloatLiteralType`, or using explicit types directly).

tgrapperon: > actually the current implementation would fail for negative Ints too.
Good catch!
As I'm getting it, it would work using `.description`, no? I didn't have the time to check out the branch yet.

tgrapperon: @iampatbrown Ok, I've just checked, and reverse sorting lexically by `.description` seems to do the trick too, but the `toPrefix` seems more efficient/elegant. It doesn't seem to cause issues with same-length strings because they're always different numbers(e.g., "-4" vs "44").

iampatbrown: Well the sort happens inside the init of the parser where `RawValue` has no constraints apart from being comparable. The `LosslessStringConvertible` was only for a static initilizer. I think you're right about using length though. The sort is to ensure the parser tries to match longer cases first. For `RawValue == String` using `Comparable` works fine.

tgrapperon: @iampatbrown That's a chance you did spot the negative `Int` case. This is a very common situation!

iampatbrown:
@stephencelis @mbrandonw Here is a summary of the possible changes from the discussion @tgrapperon and I had:

1. Sort by `endIndex` instead of `rawValue` and drop `Comparable` requirement. `Comparable` works for `String` but not for `Int` because negative numbers will not get sorted by descending length. 

```swift
self.cases = Output.allCases
  .sorted(by: { toPrefix($0.rawValue).endIndex > toPrefix($1.rawValue).endIndex })
```

2. `RawValue: LosslessStringConvertible` could be used in place of `RawValue == Int`. It would add support for additional types and would work with the current  `RawValue == Int` implementation (if sorted by length)

I've included the changes in the below branch if needed.


https://github.com/pointfreeco/swift-parsing/compare/case-iterable-raw-representable...iampatbrown:case-iterable-raw-representable

tgrapperon:
@iampatbrown Since `StringProtocol: LosslessStringConvertible`, aren't we be able to drop the `RawValue: String` overload too?

iampatbrown:
@tgrapperon Yep. It could be dropped. The main reason I didn't suggest it was the redundant call to `String.init` in `toPrefix: { String($0) }`. That's probably insignificant though.

tgrapperon:
Wouldn't `toPrefix: { $0 is String ? $0 : String($0) }` work? (but I feel we're entering the bikeshedding zone).

iampatbrown:
Using `toPrefix: { String($0) }` would be fine and would probably be the same as `toPrefix: { $0 }` in this case.

iampatbrown:
@tgrapperon for some additional context where is the implementation from [swift/String.swift](https://github.com/apple/swift/blob/ba149d964ff49ec9f72090ea9a5f7ded35abdfa2/stdlib/public/core/String.swift#L896-L911)

```swift
  ...
  /// Creates an instance from the description of a given
  /// `LosslessStringConvertible` instance.
  @inlinable @inline(__always)
  public init<T: LosslessStringConvertible>(_ value: T) {
    self = value.description
  }
}

extension String: CustomStringConvertible {
  /// The value of this string.
  ///
  /// Using this property directly is discouraged. Instead, use simple
  /// assignment to create a new constant or variable equal to this string.
  @inlinable
  public var description: String { return self }
}
```

I'd say using the lossless initializer for `String` would pretty much be the same. So I don't see anything wrong with dropping `RawValue == String` as well.

tgrapperon:
From [String.swift:L899](https://github.com/apple/swift/blob/ba149d964ff49ec9f72090ea9a5f7ded35abdfa2/stdlib/public/core/String.swift#L899) and below, it is as optimal as it can be

iampatbrown:
I'll add an edit to the summary.

tgrapperon:
As `value.description` is hit anyway, I wonder if we couldn't use this directly instead of going through `toPrefix` to sort the cases. 
Are there representations where the ordering could be different, that is where `a.description.endIndex < b.description.endIndex` doesn't implies that `toPrefix(String(a)).endIndex < `toPrefix(String(b)).endIndex`?
This is mostly by curiosity, and going through `toPrefix` is definitely a safe and acceptable solution.

I also wonder up to which point comparing indices from different `Strings` gives a meaningful result. I vaguely remember someone from the core team pointing this, but I can't figure out where yet, nor if it ever happened. It would make sense though.
In any case, we can always use the distance from `.startIndex` (that is `.count`), but the complexity is O(n) for `String` (but we're talking about enums' `RawValue`s, not War and Peace).

iampatbrown:
> As value.description is hit anyway, I wonder if we couldn't use this directly instead of going through toPrefix to sort the cases.

I'm not sure how sort could access `.description` unless some type info was passed to the parser init. That being said. I don't think  using `toPrefix($0.rawValue).endIndex` is 100% either. Using `Array(toPrefix($0.rawValue)).count` is probably safer. We could deliberately break `.endIndex` if we really wanted to with a custom `toPrefix`, however, the parser init is not public so right now that shouldn't happen.

tgrapperon:
If `RawValue` is `LosslessStringConvertible`, it has a `.description` that generates a `String` we can measure for sorting. But the more I think about it, and given the implementation from `LosslessStringConvertible`'s `String.init`, I don't think it makes any difference in terms of performance, as all `toPrefix` implementations are trivial. It is probably safer like you proposed, through `toPrefix`, as we then "sort like we parse". I don't have a counterexample where this would be required instead of sorting the `rawValues`'s descriptions, but it looks more resilient like this in any case.

The same argument could be used about `count`. From what I can see, `String.Index` comparison is only meaningful in the same `String`. They internally sort using some integer called `transcodedIndex`, which is stored along another integer called `encodedIndex`. I don't really grasp the correspondence with the internal representation, but I guess it's safer to assume that `transcodedIndex` is string-dependent (I would still like to find a reference to this). In this specific use case, I can't see how this could lead to two `rawValue`s having the same prefix and a wrong ordering, but it's most likely a bad practice in the general case. For this reason, I think it's worth the small expense of sorting using `.count`. The current implementation of the PR is using `Comparable` which is not better in terms of complexity.

iampatbrown:
> If RawValue is LosslessStringConvertible, it has a .description that generates a String we can measure for sorting.

Just want to clarify, right now the sorting occurs inside the init of `CaseIterableRawRepresentableParser` where there is no `LosslessStringConvertible` constraint. That's the reason for using `toPrefix` and why I keep saying `.description` can't be used (without making changes to the init). 

> From what I can see, String.Index comparison is only meaningful in the same String

I'm not exactly sure of the inner workings of `String.Index` but the following from [StringIndex.swift](https://github.com/apple/swift/blob/ba149d964ff49ec9f72090ea9a5f7ded35abdfa2/stdlib/public/core/StringIndex.swift#L282-L287) makes me think it would work:
```swift
extension String.Index {
  @inlinable @inline(__always)
  internal var orderingValue: UInt64 { return _rawBits &>> 14 }
}

extension String.Index: Comparable {
  @inlinable @inline(__always)
  public static func < (lhs: String.Index, rhs: String.Index) -> Bool {
    return lhs.orderingValue < rhs.orderingValue
  }
}
```

I don't think there's any issue using count though. The actual reason I used `endIndex` in the first place was because I thought `Prefix` was a `Sequence` not a `Collection`. I don't think it matters if it's O(n), I'm pretty sure `toPrefix` is going to be O(n) for `LosslessStringConvertible` anyway (apart from when `RawValue == String`)

tgrapperon:
> Just want to clarify, right now the sorting occurs inside the init of CaseIterableRawRepresentableParser where there is no LosslessStringConvertible constraint.

Ah, sorry, I was thinking this internal parser could be constrained too without creating restrictions, as I supposed that all routes creating it are already constrained by `LosslessStringConvertible` (`String` variants were already seen as such in my mind). But this is not an issue, as sorting using simply`.description` is a bad idea (see below).

> I'm not exactly sure of the inner workings of String.Index but the following from [StringIndex.swift](https://github.com/apple/swift/blob/ba149d964ff49ec9f72090ea9a5f7ded35abdfa2/stdlib/public/core/StringIndex.swift#L282-L287) makes me think it would work

Yes, it would work in our case, but `"👨‍👨‍👧‍👧".endIndex > "AB".endIndex` shows it's not a reliable indicator to sort `String`s by length. However, I can't form an example with the same prefixes and "wrong" `endIndex`es, so `endIndex` doesn't seem to be problematic for our intent. But as you said, `.count` is not where time will be lost there.

Furthermore, we **need** to go through `toPrefix` to measure the cases:
```swift
"👨‍👨‍👧".count == "👨‍👨‍👧‍👧".count // == 1
"👨‍👨‍👧‍👧".starts(with: "👨‍👨‍👧") // false

"👨‍👨‍👧".utf8.count != "👨‍👨‍👧‍👧".utf8.count // 18 != 25
"👨‍👨‍👧‍👧".utf8.starts(with: "👨‍👨‍👧".utf8) // true
```
If we sort by `.description` which are `String, we will get ambiguous results when sorting because both have the same `count` (albeit different `endIndex`es!), and if we parse utf8 views, we may match 👨‍👨‍👧 instead of 👨‍👨‍👧‍👧. Because cases go through `toPrefix`, they're correctly ordered, so we must indeed "sort like we parse". This example answers one of my questions above.

I'm sorry for the amount of discussion over this very simple matter! It seems fine like you proposed, maybe with `.count` instead of `.endIndex` for the sake of clarity.

KeithBird (changes requested):
Fix some typos.

stephencelis:
@iampatbrown @tgrapperon Thanks for the in-depth conversation!

Brandon and I spoke about this earlier today. We definitely want to ensure negative numbers are appropriately parsed, so thanks for calling that edge case out! I've pushed a test case and fix.

As far as generalizing to `LosslessStringConvertible` goes, we agree that it would be nice, but are worried that we can't do so as simply as proposed because the formatted string of the value doesn't always have a single representation. For integers, this isn't super problematic, but take an enum that is raw-representable by a `Double`: it would need to be parsed _exactly_ as it is formatted:

```swift
enum Code: Double, CaseIterable {
  case a = 5
  ...
}

Code.parser().parse("5.0") // ✅
Code.parser().parse("5")   // ❌
```

Technically this is even a problem for `Int`s if we wanted to be lenient enough to support parsing a leading "+" sign, though this is arguably a lot less problematic:

```swift
enum Code: Int, CaseIterable {
  case a = 5
  ...
}

Code.parser().parse("5")  // ✅
Code.parser().parse("+5") // ❌
```

We think both of the above could be fixed if `Code.parser()` somehow knew to call out to `Double.parser()` and `Int.parser()` under the hood instead, which might be easier now that we unified all floating-point parsers in #177. For now, though, we think this is in good enough shape to merge.

If either of you come up with a way of supporting `Double` (or if you think you have a reasonable solution for generally parsing any `LosslessStringConvertible` values), please start a discussion or PR :smile:

mbrandonw (approved):

iampatbrown:
> As far as generalizing to `LosslessStringConvertible` goes, we agree that it would be nice, but are worried that we can't do so as simply as proposed because the formatted string of the value doesn't always have a single representation. 

@stephencelis @mbrandonw Good call. I think `String` and `Int` gives decent coverage. Do you think using  `StringProtocol` and `FixedWidthInteger` respectively would work as the constrains or does that have similar issues?

> We think both of the above could be fixed if Code.parser() somehow knew to call out to Double.parser() and Int.parser() under the hood instead

Sounds like that could require passing a custom parser that incorporates Double.parser()/Int.parser() but also checks the different representations before moving on to the next case. Interesting problem....

> we think this is in good enough shape to merge.

Agreed. Nice work!

stephencelis:
> @stephencelis @mbrandonw Good call. I think `String` and `Int` gives decent coverage. Do you think using `StringProtocol` and `FixedWidthInteger` respectively would work as the constrains or does that have similar issues?

Definitely down for those generalizations! Would you want to explore them?

PR #177, opened 2022-03-07
Parse any `BinaryFloatingPoint` type

stephencelis:
We were originally piggybacking off the family of `strto*` functions for incrementally parsing floating point numbers off of strings. This pushed us to have dedicated parsers per type.

This changed in #5, but we never thought to squash all these parsers into one by abstracting over `BinaryFloatingPoint` instead. So this PR does just that.


Sources/Parsing/Parsers/Float.swift
```diff
@@ -0,0 +1,172 @@
+extension BinaryFloatingPoint where Self: LosslessStringConvertible {
```
stephencelis, line 1: I moved this file to be called "Float.swift" now that it's generic over a "float"-named protocol. Do we want to rename "Double.md" documentation, too? Or should we be consistent and just prefer "Double" everywhere?

stephencelis, line 1: Worth noting that we need to abstract over `BinaryFloatingPoint` _and_ `LosslessStringConvertible` here. While `FixedWidthInteger` inherits from it automatically, `BinaryFloatingPoint` does not.

Sources/Parsing/Parsers/Float.swift
```diff
@@ -0,0 +1,172 @@
...
+    @inlinable
+    public init() {}
+
+    @inlinable
+    public func parse(_ input: inout Input) throws -> Output {
+      let original = input
+      let s = input.parseFloat()
+      guard let n = Output(String(decoding: s, as: UTF8.self))
+      else {
+        throw ParsingError.expectedInput("\(Output.self)".lowercased(), from: original, to: input)
```
stephencelis, line 78: Errors have changed slightly: `expected double` and `expected float` are unchanged, but if you use `Float80` you'll get `expected float80` now.

stephencelis:
Also worth noting that in abstracting we now have a `Float16.parser()` for free...

@mbrandonw `Foundation.Decimal` does _not_ inherit from `BinaryFloatingPoint`, so we still don't have a `Decimal.parser()`...is that worth adding?

Sources/Parsing/Parsers/Float.swift
stephencelis: - [ ] Remove `Double.md` and mention that `Double` works in `Float.md`
- [ ] Mention that other integer types work in `Int.md`

mbrandonw (approved):

PR #178, opened 2022-03-07
Remove `isSigned` config from `Int.parser()`

stephencelis:
We chatted about this awhile back and was fresh in mind after submitting #177.

I think it's probably a good default for this particular parser, especially since we can currently construct nonsensical parsers like `UInt.parser(isSigned: true)`.

If we need a simple numeric parser that doesn't consider the sign, maybe we should have a dedicated `Digits` parser?

### DIFF
--- a/Sources/Parsing/Documentation.docc/Articles/Design.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Design.md
@@ -79,13 +79,13 @@ to maximize readability of the parsers. For example, to parse accounting syntax 
 parenthesized numbers are negative, we can use the ``OneOf`` parser builder:
 
 ```swift
+let digits = Prefix { $0.isNumber }.compactMap(Int.init)
+
 let accountingNumber = OneOf {
-  Int.parser(isSigned: false)
+  digits
 
   Parse {
-    "("
-    Int.parser(isSigned: false)
-    ")"
+    "("; digits; ")"
   }
   .map { -$0 }
 }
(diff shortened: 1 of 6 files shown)


Sources/Parsing/Documentation.docc/Articles/Design.md
```diff
@@ -79,13 +79,13 @@ to maximize readability of the parsers. For example, to parse accounting syntax
 parenthesized numbers are negative, we can use the ``OneOf`` parser builder:
 
 ```swift
+let digits = Prefix { $0.isNumber }.compactMap(Int.init)
```
stephencelis, line 82:
```suggestion
let digits = Prefix { $0 >= "0" && $0 <= "9" }.compactMap(Int.init)
```

Sources/Parsing/Documentation.docc/Articles/ErrorMessages.md
```diff
@@ -30,13 +30,13 @@ a parser that can parse accounting style of numbers, i.e. plain numbers are cons
 and numbers in parentheses are considered negative:
 
 ```swift
+let digits = Prefix { $0.isNumber }.compactMap(Int.init)
```
stephencelis, line 33:
```suggestion
let digits = Prefix { $0 >= "0" && $0 <= "9" }.compactMap(Int.init)
```

mbrandonw (approved):

PR #185, opened 2022-03-21
Whitespace(.horizontal) and Whitespace(.vertical)

stephencelis:
This changes `Whitespace` to be a bit more reasonable to use:

- It can now be configured to consume only horizontal _or_ vertical whitespace.
- It is now Unicode friendly
- It is now `Void`-returning (trivia can be useful, but isn't a great default; we'll have a solution to this breaking change in a future PR)

### DIFF
--- a/Sources/Parsing/Parsers/Whitespace.swift
+++ b/Sources/Parsing/Parsers/Whitespace.swift
@@ -1,18 +1,18 @@
-/// A parser that consumes all ASCII whitespace from the beginning of the input.
-///
-/// - Note: This parser only consumes ASCII spaces (`" "`), newlines (`"\n"` and `"\r"`), and tabs
-///   (`"\t"`). If you need richer support that covers all unicode whitespace, use a ``Prefix``
-///   parser that operates on the `Substring` level with a predicate that consumes whitespace:
-///
-///   ```swift
-///   Prefix { $0.isWhitespace }
-///   ```
+/// A parser that consumes whitespace from the beginning of input.
 public struct Whitespace<Input: Collection, Bytes: Collection>: Parser
 where
   Input.SubSequence == Input,
-  Bytes.SubSequence == Bytes,
-  Bytes.Element == UTF8.CodeUnit
+  Bytes.Element == UTF8.CodeUnit,
+  Bytes.SubSequence == Bytes
 {
+  public enum Configuration {
+    case all
+    case horizontal
+    case vertical
+  }
+
+  public let configuration: Configuration
+
   @usableFromInline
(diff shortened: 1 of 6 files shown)


Sources/Parsing/Parsers/Whitespace.swift
```diff
@@ -1,64 +1,132 @@
...
-      byte == .init(ascii: " ")
-        || byte == .init(ascii: "\n")
-        || byte == .init(ascii: "\r")
-        || byte == .init(ascii: "\t")
-    })
-    input.removeFirst(output.count)
-    return self.fromBytes(output)
-  }
-}
+  public func parse(_ input: inout Input) {
```
stephencelis, line 23: There may be better and more performant ways of defining the below, but it seems straightforward and doesn't seem to have a huge impact on existing performance (it is slightly slower, though)

stephencelis:
I guess one open question:

Should we always handle unicode whitespace (as the PR currently is implemented)? Should we only check ASCII when working on `Substring.UTF8View`?

tgrapperon:
Is "vertical" the correct metaphor for newlines? I had to think twice when discovering the PR as it made me think about horizontal/vertical selection first. With `CharacterSet`, we already have `whitespaces`, `newlines` and `whitespacesAndNewlines`. Shouldn't the API mimic this instead?

tgrapperon:
Thanks for the feedback @stephencelis. I was merely suggesting using `.whitespaces` and `.newlines` in place of `.horizontal` and `.vertical`. I've also heard about horizontal and vertical spacing in typography, but vertical whitespace is more exotic I guess.

Do you thing this Parser should be extended to support other spacing chars matching what's happening in CharacterSet like `\v` (the vertical tab) for example, or is it better to wait for the new `String`'s types and mirror what's happening there?

stephencelis:
> Do you thing this Parser should be extended to support other spacing chars matching what's happening in `CharacterSet` like `\v` (the vertical tab) for example, or is it better to wait for the new `String`'s types and mirror what's happening there?

I _think_ that's covered in the new `parse` logic, right?

One thing we've explored in the past and need to revisit is further generalizing `Prefix` over a `Predicate` protocol so that we could theoretically conform `CharacterSet` (and the forthcoming `CharacterClass`) to it:

```swift
Prefix(1..., .whitespaces)
```

Another benefit of protocolizing the predicate is we could theoretically make `Prefix` conditionally equatable, codable, etc.

stephencelis:
> Do you want me to prepare a PR for these values?

Ah interesting, yeah I overlooked those line break properties while consulting the spec. Please do!

PR #189, opened 2022-03-22
Complete Whitespace cases

tgrapperon:
Following #185, I've completed the list of cases to match `CharacterSet.whitespace` and `CharacterSet.newlines`.
I've added 1 horizontal case (`U+1680`).
I've added 3 vertical cases (`U+000B`, `U+000C`, and `U+0085`).
For consistency, I've replaced `.init(ascii:)` occurrences with their corresponding hexadecimal literal value. I've also used hexadecimal literals instead of decimal when they match UTF-8 codepoints. I can switch to all decimal/hexadecimal if preferable.
I've also added a few comments to improve readability. 
I can reverse these last changes if needed.
I didn't add new tests as the current implementation isn't exhaustive either. I can add a few ones if needed.

### DIFF
--- a/Sources/Parsing/Parsers/Whitespace.swift
+++ b/Sources/Parsing/Parsers/Whitespace.swift
@@ -26,19 +26,30 @@ where
 
     @inline(__always)
     func consumeHorizontal() -> Bool {
+      // Unicode chars from `CharacterSet.whitespaces`
+      // General category Zs + \t
       switch bytes.first {
-      case .init(ascii: " "), .init(ascii: "\t"):
+      case 0x20, 0x9: // U+0020, \t (U+0009)
         bytes.removeFirst()
         return true
 
-      case 194:
-        if bytes.dropFirst().first == 160 {
+      case 194: // U+00A0
+        if bytes.dropFirst().first == 0xA0 {
           bytes.removeFirst(2)
           return true
         }
         return false
+        
+      case 225: // U+1680
+        if bytes.dropFirst().first == 154,
+          bytes.dropFirst(2).first == 128
+        {
+          bytes.removeFirst(3)
+          return true
+        }
+        return false
 
-      case 226:
+      case 226: // U+2000 ~ U+200A, U+202F
         switch bytes.dropFirst().first {
         case 128:
           if let byte = bytes.dropFirst(2).first,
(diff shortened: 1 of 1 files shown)


Sources/Parsing/Parsers/Whitespace.swift
```diff
@@ -26,19 +26,30 @@ where
 
     @inline(__always)
     func consumeHorizontal() -> Bool {
+      // Unicode chars from `CharacterSet.whitespaces`
+      // General category Zs + \t
       switch bytes.first {
-      case .init(ascii: " "), .init(ascii: "\t"):
+      case 0x20, 0x9: // U+0020, \t (U+0009)
```
stephencelis, line 32: FWIW we haven't found there to be much of a difference in using literals vs. this initializer, but if you've found to the contrary it could be worth it!

Sources/Parsing/Parsers/Whitespace.swift
```diff
@@ -74,12 +85,20 @@ where
 
     @inline(__always)
     func consumeVertical() -> Bool {
+      // Unicode chars from `CharacterSet.newlines`
       switch bytes.first {
-      case .init(ascii: "\n"), .init(ascii: "\r"):
+      case 0xA, 0xB, 0xC, 0xD: // U+000A ~ U+000D
```
stephencelis, line 90:
```suggestion
      case 0xA ... 0xD: // U+000A ~ U+000D
```

stephencelis (approved):
Thanks!

tgrapperon:
Ah, I ran the tests, but not the benchmarks. I'll look into it.

Sources/Parsing/Parsers/Whitespace.swift
tgrapperon: This change makes the benchmarks fail to build. `bytes.first` is an optional and Swift can't match them with a `ClosedRange<UInt>`. I think its maybe simpler to revert the change. I've just checked that it build and runs the benchmarks without issues after that.

Sources/Parsing/Parsers/Whitespace.swift
```diff
@@ -74,12 +85,20 @@ where
 
     @inline(__always)
     func consumeVertical() -> Bool {
+      // Unicode chars from `CharacterSet.newlines`
       switch bytes.first {
-      case .init(ascii: "\n"), .init(ascii: "\r"):
+      case 0xA ... 0xD: // U+000A ~ U+000D
```
stephencelis, line 90:
```suggestion
      case 0xA, 0xB, 0xC, 0xD: // U+000A ~ U+000D
```

PR #191, opened 2022-03-25
Require pipes over collections of input to fully consume input

stephencelis:
I was surprised to find a parser was succeeding because a parser pipe was happily discarding trailing input.

Calling the non-inout version of `.parse(string)` inserts an `End` parser automatically, and I think we should have similar behavior here. So this PR appends a `PipeEnd` parser to the end of a parser pipe when the input is a collection. `PipeEnd` is basically an `End` parser with custom error behavior.

### DIFF
--- a/Sources/Parsing/Parsers/Pipe.swift
+++ b/Sources/Parsing/Parsers/Pipe.swift
@@ -69,8 +69,8 @@ extension Parsers {
   }
 
   public struct PipeEnd<Input: Collection>: Parser {
-    @inlinable
-    public init() {}
+    @usableFromInline
+    init() {}
 
     @inlinable
     public func parse(_ input: inout Input) throws {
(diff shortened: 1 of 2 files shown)


iampatbrown:
Good catch. 

Maybe this is an opportunity to introduce some sort of `.mapError(_:)` so `PipeEnd` could be `End().mapError(???)`. 

Could also switch out `PipeEnd` in the future too. Just wasn't too sure about the public `.init` do you think it will ever be constructed outside of `Pipe`?

stephencelis:
> Could also switch out `PipeEnd` in the future too. Just wasn't too sure about the public `.init` do you think it will ever be constructed outside of `Pipe`?

Should probably be private before merging

- [ ] Make `public init` a `fileprivate` one

I'm hoping parsers like this can all get the `some` treatment in 5.7 :smile:

iampatbrown:
> Should probably be private before merging

Yeah that's what I was thinking.

5.7 can't come soon enough! :)

mbrandonw (approved):

PR #192, opened 2022-03-28
RFC: Use `buildExpression` to promote substring parsers to UTF-8

stephencelis:
This PR introduces substantial changes to parser builder block behavior, which may not be quite right yet, but we think it's a _step_ in the right direction to make the library much more approachable.

First, it adds an empty `ParserBuilder.buildBlock()`, so that this is now valid:

```swift
let parser = Parse {}
```

This defaults to an `Always<Substring.UTF8View, Void>` parser. A small win, but friendlier.

Second, and more substantially, it adds `buildExpression`s to both `ParserBuilder` and `OneOfBuilder`, defined such that when a `Substring` parser is encountered, it automatically converts it to a `Substring.UTF8View` parser by wrapping it in `FromSubstring`.

This allows you to work with parsers on either layer freely without needing to explicitly use `FromUTF8` or `FromSubstring`:

```swift
let parser = Parse {
  "caf".utf8  // parse quickly from UTF-8
  "é"         // parse all representations of é
}
```

This change also allows us to simplify a bunch of parsers. We can get rid of a bunch of interfaces that were there to simply wrap a parser in a `FromSubstring` (`Int.parser()`, `Double.parser()`, etc.), and we can get rid of an extra generic on parsers that only needed them to move between the parser's input type and the underlying bytes (`Whitespace`, `Digits`, etc.), and a bunch of simple parsers that were previously ambiguous now compile.

For example, @haikusw and @andreweades have [pointed out](https://github.com/pointfreeco/swift-parsing/discussions/155) that library newcomers quickly meet compiler error walls when first dabbling. Simple parsers where the `Input` is not known currently fail to compile:

```swift
Parse { // 🛑 Type of expression is ambiguous without more context
  Int.parser()
}

Parse { // 🛑 Type of expression is ambiguous without more context
  Whitespace()
}
```

And the only workaround thus far has been to make the input explicit, which is not ideal:

```swift
Parse {
  Int.parser(of: Substring.self)
}
// or
Parse {
  ""
  Int.parser()
}
```

But with this PR's changes these simple cases now compile just fine:

```swift
Parse {
  Int.parser()
}
Parse {
  Whitespace()
}
```

While all of these results are promising, it does force you to rethink what _kind_ of a parser a parser is. It's no longer correct to look at a `Parse` block and think "this is a substring parser" or "this is a UTF-8 parser". Instead, it's a UTF-8 parser by default (unless you're working with an entirely different input type).

Before, it was common to work on the substring level and let the library automatically dip down into UTF-8 where appropriate, but this PR kind of "inverts" things. The end call sight should look the same, though (`parser.parse(string)`), so we're not sure if this change is negative. (We could make the default layer `Substring` instead, and `buildExpression` could automatically wrap UTF-8 parsers in `FromUTF8`, but we were worried this would negatively impact performance...this may be worth actually benchmarking, though, as it's probably a slightly better end-user experience.)

This change is potentially quite impactful, so we'd like to measure how it affects others before merging. I'd like to tag a few more active library users to get their opinion on the change, and maybe see if it negatively impacts any of their projects:

- @finestructure
- @iampatbrown
- @MaxDesiatov
- @randomeizer 
- @tgrapperon

### DIFF
--- a/Sources/Parsing/Builders/OneOfBuilder.swift
+++ b/Sources/Parsing/Builders/OneOfBuilder.swift
@@ -18,6 +18,20 @@
 /// ```
 @resultBuilder
 public enum OneOfBuilder {
+  /// Provides support for specifying a parser in ``OneOfBuilder`` blocks.
+  @inlinable
+  public static func buildExpression<P: Parser>(_ parser: P) -> P {
+    parser
+  }
+
+  /// Provides support for specifying a substring parser in a UTF-8 ``OneOfBuilder`` block.
+  @inlinable
+  public static func buildExpression<P>(
+    _ parser: P
+  ) -> FromSubstring<Substring.UTF8View, P> {
+    FromSubstring(substringParser: parser, toSubstring: Substring.init, fromSubstring: { $0.utf8 })
+  }
+
   /// Provides support for `for`-`in` loops in ``OneOfBuilder`` blocks.
   ///
   /// Useful for building up a parser from a dynamic source, like for a case-iterable enum:
--- a/Sources/swift-parsing-benchmark/XCTestLogs.swift
+++ b/Sources/swift-parsing-benchmark/XCTestLogs.swift
@@ -50,7 +50,7 @@ private let testCaseBody = Parse {
 struct TestCaseBody: Parser {
   func parse(
     _ input: inout Substring.UTF8View
-  ) throws -> (file: Substring.UTF8View, line: Int, message: Substring.UTF8View) {
+  ) throws -> (file: Substring.UTF8View, line: Int, message: Substring) {
     guard input.first == .init(ascii: "/")
     else { throw ParsingError() }
 
@@ -98,7 +98,7 @@ private let testFailed = Parse {
 .map { testName, bodyData, time in
   bodyData.map { body in
     TestResult.failed(
-      failureMessage: Substring(body.2),
+      failureMessage: body.2,
       file: Substring(body.0),
       line: body.1,
       testName: Substring(testName),
--- a/Tests/ParsingTests/ConditionalTests.swift
+++ b/Tests/ParsingTests/ConditionalTests.swift
@@ -13,13 +13,13 @@ final class ConditionalTests: XCTestCase {
 
   func testFirst() {
     var input = "42 Hello, world!"[...]
-    XCTAssertEqual(true, try parser.parse(&input))
+    XCTAssertEqual(true, try parser.parse(&input.utf8))
     XCTAssertEqual(" Hello, world!", input)
   }
 
   func testSecond() {
     var input = "43 Hello, world!"[...]
-    XCTAssertThrowsError(try parser.parse(&input)) { error in
+    XCTAssertThrowsError(try parser.parse(&input.utf8)) { error in
       XCTAssertEqual(
         """
         error: OddNumberError()
--- a/Sources/Parsing/Builders/ParserBuilder.swift
+++ b/Sources/Parsing/Builders/ParserBuilder.swift
@@ -14,6 +14,26 @@
 /// ```
 @resultBuilder
 public enum ParserBuilder {
+  /// Provides support for specifying an "empty" parser in ``ParserBuilder`` blocks.
+  @inlinable
+  public static func buildBlock() -> Always<Substring.UTF8View, Void> {
+    .init(())
+  }
+
+  /// Provides support for specifying a parser in ``ParserBuilder`` blocks.
+  @inlinable
+  public static func buildExpression<P: Parser>(_ parser: P) -> P {
+    parser
+  }
+
+  /// Provides support for specifying a substring parser in a UTF-8 ``ParserBuilder`` block.
+  @inlinable
+  public static func buildExpression<P>(
+    _ parser: P
+  ) -> FromSubstring<Substring.UTF8View, P> {
+    FromSubstring(substringParser: parser, toSubstring: Substring.init, fromSubstring: { $0.utf8 })
+  }
+
   /// Provides support for specifying a parser in ``ParserBuilder`` blocks.
   @inlinable
   public static func buildBlock<P: Parser>(_ parser: P) -> P {
--- a/Sources/Parsing/Parsers/Bool.swift
+++ b/Sources/Parsing/Parsers/Bool.swift
@@ -30,22 +30,6 @@ extension Bool {
   ) -> Parsers.BoolParser<Substring.UTF8View> {
     .init()
   }
-
-  /// A parser that consumes a Boolean value from the beginning of a substring.
-  ///
-  /// This overload is provided to allow the `Input` generic to be inferred when it is `Substring`.
-  ///
-  /// See <doc:Bool> for more information about this parser.
-  ///
-  /// - Parameter inputType: The `Substring` type. This parameter is included to mirror the
-  ///   interface that parses any collection of UTF-8 code units.
-  /// - Returns: A parser that consumes a Boolean value from the beginning of a substring.
-  @inlinable
-  public static func parser(
-    of inputType: Substring.Type = Substring.self
-  ) -> FromUTF8View<Substring, Parsers.BoolParser<Substring.UTF8View>> {
-    .init { Parsers.BoolParser<Substring.UTF8View>() }
-  }
 }
 
 extension Parsers {
--- a/Sources/Parsing/Parsers/Digits.swift
+++ b/Sources/Parsing/Parsers/Digits.swift
@@ -22,11 +22,10 @@
 /// }
 /// .parse("20220131")  // Date(year: 2022, month: 1, day: 31)
 /// ```
-public struct Digits<Input: Collection, Bytes: Collection>: Parser
+public struct Digits<Input: Collection>: Parser
 where
   Input.SubSequence == Input,
-  Bytes.Element == UTF8.CodeUnit,
-  Bytes.SubSequence == Bytes
+  Input.Element == UTF8.CodeUnit
 {
   @usableFromInline
   let maximum: Int?
@@ -35,30 +34,20 @@ where
   let minimum: Int
 
   @usableFromInline
-  let toBytes: (Input) -> Bytes
-
-  @usableFromInline
-  let fromBytes: (Bytes) -> Input
-
-  @usableFromInline
-  init<R: CountingRange>(
-    length: R,
-    toBytes: @escaping (Input) -> Bytes,
-    fromBytes: @escaping (Bytes) -> Input
-  ) {
+  init<R: CountingRange>(_ length: R) {
     precondition(length.minimum >= 1, "Can't construct Digits with length < 1")
     self.minimum = length.minimum
     self.maximum = length.maximum
-    self.toBytes = toBytes
-    self.fromBytes = fromBytes
+  }
+
+  @inlinable
+  public init() {
+    self.init(1...)
   }
 
   @inlinable
   public func parse(_ input: inout Input) throws -> Int {
-    var bytes = self.toBytes(input)
-    defer { input = self.fromBytes(bytes) }
-
-    var prefix = self.maximum.map(bytes.prefix) ?? bytes
+    var prefix = self.maximum.map(input.prefix) ?? input
     prefix = prefix.prefix(while: (.init(ascii: "0") ... .init(ascii: "9")).contains)
     let count = prefix.count
 
@@ -76,47 +65,12 @@ where
     guard let digits = Int(String(decoding: prefix, as: UTF8.self))
     else { throw ParsingError.expectedInput("digits", at: input) }
 
-    bytes.removeFirst(count)
+    input.removeFirst(count)
     return digits
   }
 }
 
-// NB: Swift 5.7 fails to build with a simpler `Bytes == Input` constraint
-extension Digits where Bytes == Input.SubSequence, Bytes.SubSequence == Input {
-  @inlinable
-  public init() {
-    self.init(1...)
-  }
-
-  @inlinable
-  public init<R: CountingRange>(_ length: R) {
-    self.init(
-      length: length,
-      toBytes: { $0 },
-      fromBytes: { $0 }
-    )
-  }
-}
-
-extension Digits where Input == Substring, Bytes == Substring.UTF8View {
-  @_disfavoredOverload
-  @inlinable
-  public init() {
-    self.init(1...)
-  }
-
-  @_disfavoredOverload
-  @inlinable
-  public init<R: CountingRange>(_ length: R) {
-    self.init(
-      length: length,
-      toBytes: { $0.utf8 },
-      fromBytes: Substring.init
-    )
-  }
-}
-
-extension Digits where Input == Substring.UTF8View, Bytes == Substring.UTF8View {
+extension Digits where Input == Substring.UTF8View {
   @_disfavoredOverload
   @inlinable
   public init() {
--- a/Sources/Parsing/Parsers/Float.swift
+++ b/Sources/Parsing/Parsers/Float.swift
@@ -32,23 +32,6 @@ extension BinaryFloatingPoint where Self: LosslessStringConvertible {
   ) -> Parsers.FloatParser<Substring.UTF8View, Self> {
     .init()
   }
-
-  /// A parser that consumes a floating-point number from the beginning of a substring.
-  ///
-  /// This overload is provided to allow the `Input` generic to be inferred when it is `Substring`.
-  ///
-  /// See <doc:Float> for more information about this parser.
-  ///
-  /// - Parameter inputType: The `Substring` type. This parameter is included to mirror the
-  ///   interface that parses any collection of UTF-8 code units.
-  /// - Returns: A parser that consumes a floating-point number from the beginning of a substring.
-  @_disfavoredOverload
-  @inlinable
-  public static func parser(
-    of inputType: Substring.Type = Substring.self
-  ) -> FromUTF8View<Substring, Parsers.FloatParser<Substring.UTF8View, Self>> {
-    .init { Parsers.FloatParser<Substring.UTF8View, Self>() }
-  }
 }
 
 extension Parsers {
--- a/Sources/Parsing/Parsers/FromSubstring.swift
+++ b/Sources/Parsing/Parsers/FromSubstring.swift
@@ -35,6 +35,17 @@ where SubstringParser.Input == Substring {
   public let toSubstring: (Input) -> Substring
   public let fromSubstring: (Substring) -> Input
 
+  @usableFromInline
+  init(
+    substringParser: SubstringParser,
+    toSubstring: @escaping (Input) -> Substring,
+    fromSubstring: @escaping (Substring) -> Input
+  ) {
+    self.substringParser = substringParser
+    self.toSubstring = toSubstring
+    self.fromSubstring = fromSubstring
+  }
+
   @inlinable
   public func parse(_ input: inout Input) rethrows -> SubstringParser.Output {
     var substring = self.toSubstring(input)
--- a/Sources/Parsing/Parsers/Int.swift
+++ b/Sources/Parsing/Parsers/Int.swift
@@ -40,28 +40,6 @@ extension FixedWidthInteger {
   ) -> Parsers.IntParser<Substring.UTF8View, Self> {
     .init(radix: radix)
   }
-
-  /// A parser that consumes an integer (with an optional leading `+` or `-` sign for signed integer
-  /// types) from the beginning of a collection of UTF-8 code units.
-  ///
-  /// This overload is provided to allow the `Input` generic to be inferred when it is `Substring`.
-  ///
-  /// See <doc:Int> for more information about this parser.
-  ///
-  /// - Parameters:
-  ///   - inputType: The `Substring` type. This parameter is included to mirror the interface that
-  ///     parses any collection of UTF-8 code units.
-  ///   - radix: The radix, or base, to use for converting text to an integer value. `radix` must be
-  ///     in the range `2...36`.
-  /// - Returns: A parser that consumes an integer from the beginning of a substring.
-  @_disfavoredOverload
-  @inlinable
-  public static func parser(
-    of inputType: Substring.Type = Substring.self,
-    radix: Int = 10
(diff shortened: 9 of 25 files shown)


Sources/Parsing/Builders/OneOfBuilder.swift
```diff
@@ -18,6 +18,20 @@
...
+  @inlinable
+  public static func buildExpression<P: Parser>(_ parser: P) -> P {
+    parser
+  }
+
+  /// Provides support for specifying a substring parser in a UTF-8 ``OneOfBuilder`` block.
+  @inlinable
+  public static func buildExpression<P>(
+    _ parser: P
+  ) -> FromSubstring<Substring.UTF8View, P> {
```
stephencelis, line 31: If this PR lands, we should probably also have a `buildExpression` for unicode scalars.

Sources/Parsing/Builders/OneOfBuilder.swift
```diff
@@ -18,6 +18,20 @@
...
+    FromSubstring(substringParser: parser, toSubstring: Substring.init, fromSubstring: { $0.utf8 })
```
stephencelis, line 32: One (unresolved) consequence of this PR is that `FromSubstring { … }` is basically not functional in builder blocks...not sure if there's a solution to this or not, but it _effectively_ doesn't matter since you can simply interleave parsers now.

Sources/swift-parsing-benchmark/XCTestLogs.swift
```diff
@@ -50,7 +50,7 @@ private let testCaseBody = Parse {
 struct TestCaseBody: Parser {
   func parse(
     _ input: inout Substring.UTF8View
-  ) throws -> (file: Substring.UTF8View, line: Int, message: Substring.UTF8View) {
+  ) throws -> (file: Substring.UTF8View, line: Int, message: Substring) {
```
stephencelis, line 53: One unintended consequence is that a `Rest` parser seems to favor the `Substring` initializer in builder blocks now...and we're not sure why yet.

Tests/ParsingTests/ConditionalTests.swift
```diff
@@ -13,13 +13,13 @@ final class ConditionalTests: XCTestCase {
 
   func testFirst() {
     var input = "42 Hello, world!"[...]
-    XCTAssertEqual(true, try parser.parse(&input))
+    XCTAssertEqual(true, try parser.parse(&input.utf8))
```
stephencelis, line 16: Another consequence of this change is that using the `inout` interfaces now goes through `.utf8`. We think this is probably OK, and uncommon outside the library's own tests, since you will usually invoke the non-`inout` version.

We could consider an overload, but we don't think it's necessary...

haikusw: Seems like it'll impact the tests of parsers build on top of the Parsing package also, won't it?  If so, that means it impacts _everyone_ who's going to be using this package, because they're all going to write copious tests, of course! ;)

finestructure:
Thanks for the heads-up, Stephen! I'll see if I can try out the changes with a one of our repos.

I've had a quick look through our Swift Package Index dependency data and these are the repos we're currently tracking as using `swift-parsing`:

- https://github.com/alexanderwe/ConventionalCommitsKit.git
- https://github.com/alexanderwe/SemanticVersioningKit.git
- https://github.com/bradhowes/swift-math-parser.git
- https://github.com/SwiftPackageIndex/ReleaseNotes.git

plus

- https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server.git

itself, which isn't in the index :)

Thought I'd share that in case you want to reach out to more users!

Tests/ParsingTests/ConditionalTests.swift
haikusw: Ok.  Remove the "seems" from that comment.  Migrated a small test project I have over to this PR branch and, yes, _all_ my tests need to be modified.  So this seems likely to impact test writing users of this package.

The rest of the changes to incorporate this PR were relatively minor.  Changing all the tests was the largest change required to migrate over (not hard of course, but seemed like busywork).

haikusw: The other place this is going to trip up new users is when they're just playing around in a playground because that's where the `inout` version of parse is often used (as you do in your videos).  Given that part of the goal of this PR is to make it easier for new users, would be better not to require they figure out to add `.utf8` (and wonder "why?").

stephencelis: > Ok. Remove the "seems" from that comment. Migrated a small test project I have over to this PR branch and, yes, _all_ my tests need to be modified. So this seems likely to impact test writing users of this package.

@haikusw Thanks for trying this branch out!

Are you asserting against the input after it's been consumed? If so, can you provide an example of why? And if not, have you tried using the non-`inout` version of `parse`?

In general we want to move in a direction that encourages using the non-`inout` version throughout the docs and tutorials.

This isn't to say we shouldn't provide an overload here should this PR land, though.

haikusw:
Checked out my various parsing test project, including some early ones were I was very confused :).

_Really_ nice to be able to replace this:

`let skippedWhitespace =  Skip<FromUTF8View<Substring, Whitespace<Substring.UTF8View>>> { FromUTF8View { Whitespace() } }`

with:

`let skippedWhitespace =  Skip { Whitespace() }`

**!!!**

As I noted next to your comment about it, having to append `.utf8` to every input passed to a `parse(&inoutInput)` was a bit annoying.  If that can be enhanced it would reduce the "huh?!" new users of this face.

stephencelis:
> _Really_ nice to be able to replace this:
> 
> `let skippedWhitespace = Skip<FromUTF8View<Substring, Whitespace<Substring.UTF8View>>> { FromUTF8View { Whitespace() } }`
> 
> with:
> 
> `let skippedWhitespace = Skip { Whitespace() }`

😄

@haikusw Another thing to note is that `Whitespace` was recently changed to be skipping/`Void` by default, since it's less common to care about the underlying "trivia" unless you're building something like SwiftSyntax, which wants to preserve that info.

(In case you _do_ want to preserve that info, there's a new `Consumed` parser that can capture the consumed input of another parser, _e.g._, `Consumed { Whitespace() }`.)

haikusw:
@stephencelis Thanks.  That does sound useful (!) and your SwiftSyntax example is a great one to keep in mind when working on this package/library.  

I actually have my own `whitespace` and `skippedWhitespace` parsers defined to match a language grammar (which, interestingly, isn't as all encompassing as the new Whitespace parser in this library 🤔), but I had that old definition above that you helped me get working sitting in a comment in a playground, so I thought I'd switch back to that and see if this PR made the 🤯 go away.  Did! 🎉

Tests/ParsingTests/ConditionalTests.swift
haikusw: Well… Mostly I was just copying your videos and playgrounds when trying out the parser library and so I wrote my tests like you guys did :)  

May also have been because of some of the non-`inout` `parse` functions weren't working on `Substring`?  I haven't had a chance to play with parsing lately, so I don't entirely recall.

Yes, I do have a couple of tests where I'm asserting that there is the expected output left after the parser runs.

I just switched all the other tests to use the non`input` version of `parse`, and they are _working fine_ now on this branch.

I took a quick scan through the referenced packages Sven listed below and I didn't see any there that require the `inout`.  It was quick, so I could have missed something, but wasn't obviously a big deal there.

So I think your sense that the `inout` version isn't needed as much appears to be spot on.

stephencelis:
I've converted this to a draft for now. While it solves real issues with the existing state of things, it introduces a layer of magic that I don't think we're comfortable with.

The core issue at hand is a [Swift result builder bug](https://bugs.swift.org/browse/SR-15403). To work around the bug, we introduce overloads for specific common inputs (like `Substring` and `Substring.UTF8View`). These overloads make unambiguous parser builder blocks work great, but introduce ambiguity for simple parser builder blocks.

So this PR is really a workaround for a problem with a workaround for a problem...and this workaround introduces its own problems! 😵

So we think the correct path forward is to maybe close this and wait for the Swift bug to be fixed, instead. Then we can remove our original workaround, which should solve the problem this workaround intended to solve :laughing:

haikusw:
Aw…. sad to revert my projects changes to try this branch, but hopefully we can get that bug fixed sooner than later.  I just upvoted it and maybe others will also.

haikusw:
Is there a simpler example that can demonstrate the bug?  I find the one in that bug report pretty brain-bending.  Just wondering if a simpler/clearer example might help get it addressed sooner.

stephencelis:
@haikusw The example is definitely abstract, but I'm not sure it can get simpler. We can make it a lil more complex but understandable by renaming a few things:

```swift
struct Parser<Input> {}

@resultBuilder
enum ParserBuilder {
  static func buildBlock<Input>(_ components: Foo<Input>...) -> Foo<Input> {
    .init()
  }
}

func parse<Input>(@ParserBuilder build: () -> Parser) -> Parser {
  build()
}

parse {
  Parser<Substring>()
  Parser() // 🛑
}
```

> error: Cannot convert value of type 'Parser<_>' to expected argument type 'Parser<Substring>'

Basically, when you're in a builder context, generics aren't inferred the same way they'd be elsewhere:

```swift
let parsers = [
  Parser<Substring>(),
  Parser() // ✅
]
```

stephencelis:
Turns out this isn't a bug, unfortunately. @iampatbrown turned my attention to this post: https://forums.swift.org/t/function-builder-cannot-infer-generic-parameters-even-though-direct-call-to-buildblock-can/35886/15

haikusw:
Thanks.  I finally ate an overly delayed breakfast and my brain restarted :)  <whew!>
Realized it was mostly naming that was making my brain tie itself in knots, and that the enum was being used as an un-creatable type as a namespace.  

Get back and that's exactly what you improved above.  Thanks.

The other thing that made me think ("why is that here?") was fully specified one.  Now I see that it's to show what the compiler needs to succeed.  Maybe a comment on that line  saying "// fully specified works, but shouldn't be necessary" or something.

Probably the Swift compiler devs are so deep in this stuff that your example is completely legible to them :)

Cheers

haikusw:
> Turns out this isn't a bug, unfortunately. @iampatbrown turned my attention to this post: https://forums.swift.org/t/function-builder-cannot-infer-generic-parameters-even-though-direct-call-to-buildblock-can/35886/15

🤔

haikusw:
That thread was interesting.  The discussion towards the end between John and Douglas and Matthew was interesting.  Not sure it applies to the issue Parsing has though.  Also sounded like it's uncharted territory in terms of compiler performance impact.

Does this mean that dropping the chaining style syntax may want to be reconsidered?
Or maybe it's that this PR gets reconsidered since that Swift "bug" is a design decision?

It's cool how much you guys do work that so often pushes up against the limits of Swift.

haikusw:
Looks like that discussion was about:

> `static func buildExpression(_ expression: Expression) -> Component`
> Builds a partial result from an expression. You can implement this method to perform preprocessing—for example, converting expressions to an internal type—**or to provide additional information for type inference at use sites.**

From: https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#ID633

haikusw:
I have no idea if this works, but it now compiles :)

```
struct Parser<Input> {}

@resultBuilder
enum ParserBuilder<Input> {
  static func buildBlock(_ parsers: Parser<Input>...) -> Parser<Input> {
    .init()
  }
  static func buildExpression(_ expression: Parser<Input>) -> Parser<Input> {
    return expression
  }
}

func parse<Input>(@ParserBuilder<Input> build: () -> Parser<Input>) -> Parser<Input> {
  build()
}

parse {
  Parser<Substring>()
  Parser()
}
```

haikusw:
Deleted comments that basically got back to what you did in this PR - use `buildExpression` like they talked about in that thread.   That thread makes it seem like the approach in this PR is what they expect one to do, especially given this from the documentation:

> `static func buildExpression(_ expression: Expression) -> Component`
> Builds a partial result from an expression. You can implement this method to perform preprocessing—for example, converting expressions to an internal type—**_or to provide additional information for type inference at use sites._**

haikusw:
> @haikusw I think `buildExpression` might be the right tool for the job here, but I don't think we're using it the right way, yet.
> 
> John McCall seems to suggest that parameterizing the builder with a generic (like `@ParserBuilder<Input>`) could be a workaround. I'm going to see how far I can take it.

Cool!

I did this to your more grokable bug sample from above:

```swift
struct Parser<Input> {}

@resultBuilder
enum ParserBuilder<Input> {
  static func buildBlock(_ parsers: Parser<Input>...) -> Parser<Input> {
    .init()
  }
  static func buildExpression(_ expression: Parser<Input>) -> Parser<Input> {
    return expression
  }
}

func parse<Input>(@ParserBuilder<Input> build: () -> Parser<Input>) -> Parser<Input> {
  build()
}

parse {
  Parser<Substring>()
  Parser()
}
```

and the compiler liked it :)
(FWTW)

nikitamounier:
This seems like a really good change to me! 

Most Swift users aren't familiar at all with all the different string flavours - I myself only discovered UTF8View, UnicodeScalarView, et. al from your video series. Juggling with all these different kinds of Strings is definitely the biggest impedance to my productivity with swift-parsing. What this change would allow us to now say is: "Mark all your strings in your parsers with .utf8 unless there are any weird characters" (basically any non-ASCII characters). I personally find that quite powerful, and a huge plus from a usability standpoint.

As you mentioned above, what we thought was a bug is unfortunately not a bug, and the proposed solution by John McCall doesn't seem optimal. 

I personally don't see this as a layer of magic, but instead as an opinion taken by the library to promote performance --- and not really at the cost of correctness since users would have needed to wrap their "weird characters" in 'FromSubstring' anyways if they were to use UTF8View anyways.

tgrapperon:
> John McCall seems to suggest that parameterizing the builder with a generic (like @ParserBuilder<Input>) could be a workaround. I'm going to see how far I can take it.

@stephencelis, I used this approach in a few of my projects/experiments with result builders, most likely for this reason. For example, the `EffectBuilder` from [TCA's #1019](https://github.com/pointfreeco/swift-composable-architecture/pull/1019) as initially generic over the `Action` and `Failure` before being nested in `Effect` to inherit these generics (with would require a bunch of `where Action == …` otherwise). I've also used one in the protocolized refactor of my `swift-measure` benchmark library which is not public yet (and where the `Experiment` protocol I introduced is very similar to `Parser`, with some `Input` and `Output`). 

I've ended up making this `ExperimentBuilder` not generic because it wasn't pairing well with the kind of call site syntax I aimed: If I recall correctly, it was similar to a chicken or egg problem, and I had to qualify the generics at the equivalent of the `Parse` level to make it compile (which I wanted the avoid). The last time I've tried, it was with Swift 5.5, but Swift 5.6 is better at resolving types so it may work better now. And maybe qualifying the `Input` of top level`Parse`s is acceptable after all, as we're already doing it frequently with their first block instead.

I kept using a parametric builder for `EffectBuilder` because this one was used in places where the generics could be inferred to match the `Reducer` expected output.

I've noticed that these generic builders are slightly less convenient to reuse, as they often require writing the generics twice in the function declaration.

This is mostly my personal recollection of previous experiments with parametric builders. I'll give it a new spin with `ParserBuilder`s to check if it improved or something new clicks.

It is also possible that the new `Reducing` blocks likely expected in Swift 5.7 can support the generic propagation between blocks, but I didn't experiment with them yet.

I'm also remembering that result builders can be defined using protocols. This allows to compose/specialize behaviors, so maybe a `UTF8ParserBuilder: ParserBuilder` could help solving the PR's issue, but this would require many overloads in built-in parser's definitions.

stephencelis:
@tgrapperon I have a branch (unpushed) that is both promising and frustrating. It solves the generic inference issues but I believe the generic and/or `buildExpression`s introduce regressions in type-checking that make something like the user CSV parser fail to parse in time (due to the overload space of `String.init` and `.map(String.init)`).

I'll try to push something that works but has had to refactor around these regressions tomorrow, if you're interested in taking a stab at things.

> It is also possible that the new `Reducing` blocks likely expected in Swift 5.7 can support the generic propagation between blocks, but I didn't experiment with them yet.

While this could potentially be true (according to evolution lore), it didn't appear to be the case in my brief testing. I commented in the proposal review today: https://forums.swift.org/t/se-0348-buildpartialblock-for-result-builders/56143/24

randomeizer:
I will have a play with it in a branch of hsdoc shortly. I need to update it to the latest `main` anyway, will start with that and then play here.

Having not looked at the code, I'm ok in principle with trying to make it easier to use the more efficient `utf8` approach. Like others here, this series was the first encounter I had with it, and it's not necessarily intuitively clear that a UTF-8 collection will not necessarily be a 1-to-1 byte count with the number of characters that end up on your screen. I get it, but largely because I've written a UTF-8/UTF-16 parser (in a different language) in the past.

In some ways, it's the layers/alternative ways of dealing with different abstractions of character strings that makes the library more complicated. Not much that can be done about that, particularly if writing custom parsers. It is probably something that's worth writing a documentation page about specifically. Going from `String` to `Substring` to a collection of UTF characters. (This may already exist, I haven't had a chance to review the new DocC articles yet)

Anyway, will update when I get a chance to kick the tires.

Sources/swift-parsing-benchmark/XCTestLogs.swift
iampatbrown: @stephencelis you'd probably need to disfavor the `buildExpression` that returns `FromSubstring`. For me it was taking precedence over 

```swift
@inlinable
public static func buildExpression<P: Parser>(_ parser: P) -> P {
  parser
}
```

tgrapperon:
> both promising and frustrating

@stephencelis This is the experience I had, and why I finally concluded at the time that "the less generic/specialized the builder, the better". But it may be possible to reach something nice with more effort than I invested.

I'm traveling and I missed that `buildPartialBlock` entered official review. This is great news. I was aware of the one-way constraint behavior (albeit not the naming): I think Becca Royal Gordon mention it in her WWDC talk when writing the equivalent/intermediary representation of the blocks as individual statements, but I was hoping that the `buildPartialBlock` would provide enough structure to enforce some type along the blocks chain. It looks very unlikely from what I'm understanding now.

I have some wild ideas to check, but I'll experiment more before drowning you with nonsense.

stephencelis:
@tgrapperon I spun up a `build-expression-2` branch that introduces a generic to `ParserBuilder` and `OneOfBuilder`, but type-checking gets bogged down quickly to the point of the library being unusable.

You can see this by running `OneOfTests.testJSON`, where this expression takes 5 seconds to type check for me:

https://github.com/pointfreeco/swift-parsing/blob/build-expression-2/Tests/ParsingTests/OneOfTests.swift#L212-L222

(10 seconds with the final parser commented in.)

Whereas on `main` it builds pretty much instantaneously.

The `string` parser just below fails to compile after 60 seconds of trying.

So I'm not too confident this approach is going to work, unfortunately. I thought maybe the many generated `Zip`s was making things worse, but when I tested the simpler `buildPartialBlock` approach, type-checking speeds were more or less the same (awful).

Putting `ParserBuilder<Input>` aside for a moment, I don't even think the enhanced `buildPartialBlock` with better type inference would help us here, since one-way constraints would still mean that the first parser in a builder block would need to build in isolation.

```swift
Parse {  // What is this parser's `Input`?
  Int.parser()
}
```

So I don't think that feature would help us, either.

In the end I'm wondering if the best we're going to get is when we can define parsers in a SwiftUI-like fashion:

```swift
struct QuotedField: ParserPrinter {
  var body: some ParserPrinter<Substring, Substring> {
    "\""
    Prefix { $0 != "\"" }
    "\""
  }
}
```

If the default way to build a parser involves the ceremony of introducing a type and scope where the `Input` is fully-determined, the opportunity for ambiguity disappears.

stephencelis:
After a bunch more experimentation, we've surfaced what appear to be bugs with parameterized result builders that are blocking:

- [SR-16088: Parameterized result builder breaks type inference](https://bugs.swift.org/browse/SR-16088)
- [SR-16090: Nested parameterized result builders are slow](https://bugs.swift.org/browse/SR-16090)

Should these issues be fixed, there's could be a nice inference story for the future of swift-parsing via a SwiftUI-like interface, where `Input` is always made explicit.

If anyone is interested in seeing them fixed, please feel free to upvote them on JIRA.

stephencelis:
@tgrapperon Thanks for exploring and letting us know! I'm not quite sure if separating the parser from the DSL is possible, or if it would solve the type inference issues we are seeing, but if you come up with anything interesting we'd love to see it!

stephencelis:
I've also posted to the forums to ask about the issues we were seeing with generic result builders: https://forums.swift.org/t/result-builder-expressiveness-and-type-inference/56417

haikusw:
> After a bunch more experimentation, we've surfaced what appear to be bugs with parameterized result builders that are blocking:
> 
> * [SR-16088: Parameterized result builder breaks type inference](https://bugs.swift.org/browse/SR-16088)
> * [SR-16090: Nested parameterized result builders are slow](https://bugs.swift.org/browse/SR-16090)
> 
> Should these issues be fixed, there's could be a nice inference story for the future of swift-parsing via a SwiftUI-like interface, where `Input` is always made explicit.
> 
> If anyone is interested in seeing them fixed, please feel free to upvote them on JIRA.

Thanks!  Upvoted.

Curious about the gist you [posted](https://gist.github.com/stephencelis/bebaae763aa46fdf640bf6ba1f3226d1#file-parameterizedsluggishness-swift-L32)

I don't have the current snapshot with the experimental feature, but found the [pitch](https://forums.swift.org/t/pitch-buildpartialblock-for-result-builders/55561/10).  Sounds like a comment in that pitch about these two bugs would be apropos? Especially since you were so enthusiastic in your comment there :)

haikusw:
Well so much for hoping it was in the WIP implementation of `buildPartialBlock` support and thus easier to get fixed… :-/

Thanks.

haikusw:
It's uncanny how often you guys push the Swift compiler and either find bugs or language limitations… 🧙‍♂️

stephencelis:
> It's uncanny how often you guys push the Swift compiler and either find bugs or language limitations… 🧙‍♂️

Wonder if Apple should hire us to QA 😆

stephencelis:
Closing this for now. It was an interesting experiment but we're not going to commit to such change right now :smile:

PR #200, opened 2022-04-19
Fix overflow when comparing non-alphabetical chars

### LINKED ISSUE #199: Double parser crashes when parsing "…" character
When parsing a `Double` from a `Substring` that starts with a special ellipsis character, the program crashes.

Test to reproduce:

```swift
func testEllipse() throws {
        var test = "…"[...]
        _ = try Double.parser().parse(&test)
}
```

Console: `Thread 1: Swift runtime failure: arithmetic overflow`

image
image

tgrapperon:
This should fix #199.
The char range is checked to be in [A-Z] (that is `65...90`) before trying to lowercase the char by adding 32. 
This prevents going beyond `UInt8.max` by mistake.
I added a test that passed, but I removed it as it was only testing some internal method implementation.


stephencelis (approved):
Thanks for the fix, @tgrapperon!

PR #217, opened 2022-06-08
Fix typo in `flatMap` documentation

stephencelis:
Confusing one!


Sources/Parsing/ParserPrinters/Many.swift
```diff
@@ -303,7 +303,7 @@ extension Many where Printability == Never {
     @ParserBuilder element: () -> Element,
     @ParserBuilder separator: () -> Separator,
     @ParserBuilder terminator: () -> Terminator
-  ) where Printability == Never {
```
stephencelis, line 306: These redundant warnings show up in Xcode 14.

mbrandonw (approved):
Oops!

lukeredpath:
@stephencelis thanks for fixing this. Would it be possible for the documentation to explain how you would use map with a conversion to replicate flatMap functionality? For instance, I can see there’s a parser conversion [here](https://github.com/pointfreeco/swift-parsing/blob/main/Sources/Parsing/Conversions/ConversionParser.swift) but it’s not clear how to use it (and it’s documentation seems to reference a none existence Conversion/parse method, or I couldn’t find it).

stephencelis:
> For instance, I can see there’s a parser conversion [here](https://github.com/pointfreeco/swift-parsing/blob/main/Sources/Parsing/Conversions/ConversionParser.swift) but it’s not clear how to use it (and it’s documentation seems to reference a non-existent Conversion/parse method, or I couldn’t find it).

@lukeredpath That conversion's actually a bit confusing, and I thought we'd removed it before merging the printer branch. I've just PR'd fully removing it here: #219.

> Would it be possible for the documentation to explain how you would use map with a conversion to replicate flatMap functionality?

We'll have to revisit docs when we have a moment, but if you provide an example usage of `flatMap` we can try to provide a refactoring that uses `map(_:Conversion)` instead.

lukeredpath:
👍🏻 I've started a discussion here https://github.com/pointfreeco/swift-parsing/discussions/220.

PR #225, opened 2022-06-10
Add CheckTaskCancellation

tgrapperon:
Following the discussion #221, this PR proposes to add a trivial `CheckTaskCancellation` parser/printer that can early-out parsing if the parent task is canceled. It is used as one would use `try Task.checkCancellation()`, that is before a potentially lengthy operation:
```swift
Many {
  CheckTaskCancellation()
  Parse {…}
}
```
This gives the user the possibility to perform these checks where and when it is relevant.

I find the enclosing or fluent alternative discussed in #221 more difficult to understand because of the parser they should be applied to. Users using `Tasks` should likely be familiar with this PR's construct.

This doesn't fix the issue of parsing 100 millions integers. In this case, we would only want to perform the checks every 10000 integers for example, but the `CheckTaskCancellation` style doesn't allow this. It is probable that we can build something that does this with existing combinators. I had a potential implementation that made `CheckTaskCancellation` a class with a counter, and some API like `CheckTaskCancellation(every: 10000)`, but I was worried about the added complexity for the user (this situation only occurs in a `Many`, but the argument is there all the time and can be used where it doesn't make sense). And there was an issue with making it `Sendable`, locks & performance. So for now, I have no solution to propose for this configuration.

In order to test it, we generate a very long string of "1,1,1,…" and ask a `Many` parser to process it in a `Task`. Quickly after the beginning of the execution, we cancel the `Task` and check how much it processed. Since the parsing was aborted, the expected count should be lower than the input's size. If we remove the check, parsing continues for a few seconds and all the input is processed. On my machine, it passes in ~0.1s, parsing ~1/10th of the input, and fails in ~4s. ~This gives some room ahead and shouldn't be too flaky~ *[Downloading Xcode from CI to check what's going on]*. There is maybe a better way to test it.

This makes me wonder if we should backtrack when we exit processing, because we have a parser that threw and consumed the input.

### DIFF
--- a/Sources/Parsing/ParserPrinters/CheckTaskCancellation.swift
+++ b/Sources/Parsing/ParserPrinters/CheckTaskCancellation.swift
@@ -1,0 +1,54 @@
+/// A trivial parser that fails if the parent task is cancelled.
+///
+/// Parsing can be a long-running procedure. For this reason, you may want to abord the execution of
+/// a parser if its parent task is cancelled. When you compose parsers using a ``ParserBuilder``,
+/// you can insert this trivial parser before a potentially time-consuming sequence of parsers to
+/// abord processing it in any case the parent task is cancelled, and free the current thread. For
+/// example:
+///
+/// ```swift
+/// let parser = Parser {
+///   Many {
+///     CheckTaskCancellation()
+///     csvLineParser
+///   }
+/// }
+/// ```
+///
+/// This parser's sole purpose is to stop processing if it is evaluated when the parent task is
+/// cancelled. It doesn't affect parsing or printing results.
+///
+/// - Throws: This parser throws a `CancellationError` if the parent task is cancelled.
+public struct CheckTaskCancellation<Input>: ParserPrinter {
+  /// Construct a parser that runs the next parsers only if the parent task is not cancelled.
+  @inlinable
+  public init() {}
+
+  @inlinable
+  func checkTaskCancellationIfNeeded() throws {
+    if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 7.0, *) {
+      #if canImport(_Concurrency) && compiler(>=5.5.2)
+      try Task.checkCancellation()
+      #endif
+    }
+  }
+
+  @inlinable
+  public func parse(_ input: inout Input) throws {
+    try checkTaskCancellationIfNeeded()
+  }
+  @inlinable
+  public func print(_ output: Void, into input: inout Input) throws {
+    try checkTaskCancellationIfNeeded()
+  }
+}
+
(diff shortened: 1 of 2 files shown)


tgrapperon:
It seems there is a bug they fixed since Xcode 13.2.1 and OS of this time. With the current code and this version of Xcode testing on iOS of this time, the task execution is blocking and `.cancel()` is called only once it's finished. If I switch to Mac, it uses my up to date Monterey runtime, the task is not blocking, cancel is called fairly quickly and the test passes. If this is a known bug, it's maybe worth adding a note in the documentation that this parser has no effect on such systems. Or we can limit the availability to these versions.

stephencelis:
@tgrapperon My main first thought at this point is to generalize. It seems like it would be useful to insert arbitrary `Void` effects into parsers in general. Spitballing syntax:

```swift
Do { try Task.checkCancellation() }
```

While the overhead of the escaping closure may not be ideal, it also may be minimal in the `body` style of parser we've been exploring.

I'm not sure we're ready to dive into the design space here before having the time to think about it and explore it more, though! Just some thoughts in the meantime :smile:

tgrapperon:
@stephencelis Thanks for the feedback. I guess it also opens the way to some kind of pre/post actions for parsers, but right now, and without some kind of `Environment`/`Context` to pass the information along, uses seem fairly limited to global stuff like Task cancellation or logging. We can capture something of course, but it would be nice if built-in parsers could pass relevant information along, like the current count in a `Many`, etc. If user-defined properties are allowed, the usual issue with environmental stuff is the cost induced by handling dictionaries, but we can maybe take some inspiration from the way `AttributedString` are composing attributes.

I'm closing the PR for now then. The failing test remains a mystery. I don't understand why it passes on my Mac but fails on an iPhone with similar OS/Swift versions, and on older macOS/Swift versions.

tgrapperon:
@stephencelis BTW, a similar pattern forms in TCA with protocolized reducers, where something that was before a wrapper of a `Reducer` can become an independent, elemental reducer that can be chained/zipped with a "base" reducer. For example, if the `Binding()` reducer is a reducer that only reduces binding actions and does nothing otherwise, the `.binding()` reducer can be rewritten as[^1]:
```swift
Reduce {
  Binding()
  Upstream()
}
```
[^1]: In the `proto-2` experiments at https://github.com/tgrapperon/swift-composable-architecture/commits/reducer-modifier, I compare both approaches.

This is very nice, because it offers the possibility to compose/break down reducers functionalities:
```swift
Reduce {
  Animation()
  Binding()
  Upstream()
  Debug()
}
```
I wonder if some existing combinators can also be broken into elemental ones.

It can be interesting to draw parallels between `swift-parsing` and TCA during your reflections around `Do {}`.

PR #229, opened 2022-06-16
[Feat] Add support for PrefixUntil

mackoj:
Hi,

This PR adds support for a parser `PrefixUntil` that behaves like `PrefixUpTo` in most cases but takes a parser too in order to validate a certain pattern.

`PrefixUntil` consumes a subsequence from the beginning of its input up to a given sequence of elements then run a parser to test them. In case of success, the prefix is returned. 

It have a fast path(🏎💨) and a slow path(🐌). The fast path is enable by providing the beginning of the parser as a string when the begining will be found it will try the real parser to confirm. The slow path consumes elements of a collection one by one until a given parser parser succeeds.

The fast path is a modified version of `PrefixUpTo` with a small twist to confirm a pattern with a parser.

`PrefixUntil` allows consuming input with a dynamic end. 

- [ ] At the moment it lack support for printing(and I have no idea how to do it if someone could help)

## Benchmark

It performs very well in most cases almost as fast as `PrefixUpTo`. On a more complex parsing challenge `FastPath` is still way faster than `SlowPath`.

```
name                                         time           std        iterations
---------------------------------------------------------------------------------
PrefixUntil.PrefixUpTo                        256875.000 ns ±  11.13 %       5191
PrefixUntil.PrefixUntil / FastPath            251291.000 ns ±  23.88 %       5554
PrefixUntil.PrefixUntil / SlowPath           4010916.000 ns ±   1.79 %        345
PrefixUntil.PrefixUntil / FastPath / Complex  425958.500 ns ±   1.76 %       3288
PrefixUntil.PrefixUntil / SlowPath / Complex 5709958.000 ns ±   2.31 %        237
```

# Links

https://github.com/pointfreeco/swift-parsing/discussions/228
https://github.com/pointfreeco/swift-parsing/discussions/223

Cheers,

### DIFF
--- a/Sources/Parsing/ParserPrinters/PrefixUntil.swift
+++ b/Sources/Parsing/ParserPrinters/PrefixUntil.swift
@@ -1,0 +1,120 @@
+import Foundation
+
+/// A parser that consumes a subsequence from the beginning of its input up to a given parser succeed.
+///
+/// This parser behave a lot like `PrefixUpTo` but it can be slower because it can use a parser instead of
+/// a static string and consume and return input up to a particular parser succeed.
+///
+/// If provided with a string that correspond to the begining of the parser it will be almost as fast as `PrefixUpTo` in most cases. The only case where it will be slow when providing a string is when there is a lot of false positive aka pattern that look like what the parser should valid but are not valid.
+///
+/// ```swift
+/// let separatorParser = Parse {
+///   "["
+///   Int.parser()
+///   Optionally {
+///     ";"
+///     Int.parser()
+///   }
+///   "m"
+/// }
+///
+/// let lineParser = PrefixUntil { separatorParser }
+///
+/// var input = "Hello[31mworld[30m\n"[...]
+/// try line.parse(&input)  // "Hello"
+/// input                   // "[31mworld[30m\n"
+/// ```
+///
+/// - Tip: This is 15x slower than `PrefixUpTo` so use it only where a dynamic limit is needed caution.
+public struct PrefixUntil<Input: Collection, Upstream: Parser>: Parser where Input.SubSequence == Input, Upstream.Input == Input {
+  
+  public let upstream: Upstream
+  public let possibleMatch: Input?
+  public let areEquivalent: (Input.Element, Input.Element) -> Bool
+  
+  @inlinable
+  public init(
+    _ possibleMatch: Input? = nil,
+    by areEquivalent: @escaping (Input.Element, Input.Element) -> Bool,
+    @ParserBuilder _ build: () -> Upstream) {
+      self.upstream = build()
+      self.possibleMatch = possibleMatch
+      self.areEquivalent = areEquivalent
+    }
+  
+  @inlinable
+  public func fastPath(_ input: inout Input) throws -> Input {
+    guard let possibleMatch = self.possibleMatch else { return input }
+    guard let first = possibleMatch.first else { return possibleMatch }
+    let count = possibleMatch.count
+    let original = input
+    
+    while let index = input.firstIndex(where: { self.areEquivalent(first, $0) }) {
+      input = input[index...]
+      if input.count >= count,
+        zip(input[index...], possibleMatch).allSatisfy(self.areEquivalent)
+      {
+        let outputSlow = try slowPath(&input)
+        if outputSlow.count == 0 { return original[..<index] }
+        return outputSlow
+      }
+      input.removeFirst()
+    }
+    input.removeFirst(original.count)
+    return original
+  }
+  
+  @inlinable
+  public func slowPath(_ input: inout Input) throws -> Input {
+    let original = input
+    var testInput = input
+    var index = input.startIndex
+    var hasParsed = (try? self.upstream.parse(&testInput)) != nil
+
+    while hasParsed == false && index != testInput.endIndex {
+      testInput.removeFirst()
+      index = testInput.startIndex
+      hasParsed = (try? self.upstream.parse(&testInput)) != nil
+    }
+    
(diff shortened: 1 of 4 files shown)


mackoj:
@tgrapperon Yes when reading your description both parsers seems to work in a similar way.

mackoj:
FWIW I don't expect this to be merge but to present an idea of a missing feature that can be recover and improve by others.

mackoj:
A way to make this parser faster is to be able to scan for the first character of subsequent parser in order to reduce the number of tests. I can play around with this optimisation and see if it become a bit more decent.

tgrapperon:
As I'm not a maintainer, I'm only providing my personal insight. It is nice that you were able to retrieve a similar functionality with acceptable performance. I'm slightly worried about the API though. This parser exhibits a very different behavior if you're able to provide it a hint or not. This is not always possible, for example when the upstream is a `OneOf`, and in this case, the parser will be very slow.
One solution would be to prune the non-hinted branch and force the user to provide the `possibleMatch` argument. This would however lower the number of potential candidates, and the API could probably be refined in this spirit. It is also possible in this case, `PrefixUntil` functionality could be built using existing parsers. For example, I feel that the following parser would be equivalent (I haven't checked, and there are maybe corner cases or better ways to achieve this):
```swift
enum LoopAction: Equatable {
  case continue
  case break
}

let parser = Consume {
  Many {
    PrefixUpTo(possibleMatch)
  } separator: {
    OneOf {
      Peek { upstream }.map { _ in LoopAction.break }
      StartsWith(possibleMatch).map { _ in LoopAction.continue }
    }.filter { $0 == .continue }
  } terminator: { 
    Peek { upstream } 
  }
}
```
If this is the case and performance is acceptable, I think that extending the library API with yet another Parser would be hard to justify. Of course, if the performance or function of the example above are not there, the fast track version of `PrefixUntil` could have its merits, maybe with another naming or API (it would be the 4th parser of the library that starts with `Prefix…`. Furthermore, the API would feel a little strange, with a redundant argument. I also wonder if this construct is generalizable or not. For example, in the parser I've presented, we could forego the whole mapping/filter thing if we could ask `Many` to check terminators before separators. I don't think it has its uses outside of this problem, but the API could have been be as nice as reordering the order of closures:
```swift
Many {} terminator: {} separator: {} // vs
Many {} separator: {} terminator: {} 
```

But again, it's my own opinion, nothing authoritative, so please don't refactor everything because of what I've said!

iampatbrown:
Hey @mackoj,

Have you considered using `Many` + `Not` as an alternative?

Actually, @tgrapperon has just echoed what I was about to say :) 

So I'll just include some examples based on your tests.

```swift
 // slow path
 output = try Consumed {
  Many(1...) {
    Not { "Hello" }
    First()
  }
}.parse(&input)
 
// fast path
output = try Consumed {
  Many(1...) {
    Not { "Hello" }
    PrefixUpTo("H")
  }
}
.parse(&input)
```
Not sure if these are going to be 1:1 like `PrefixUntil` but I think it would be possible to achieve the same result with existing parsers.

tgrapperon:
Yet again, @iampatbrown proposes a more elegant implementation than mine! ;)

mackoj:
Thanks for taking the time to respond @tgrapperon. I do agree with a lot of your points.

The API as it is can be improve. The better approach would have been to make it part of PrefixUpTo(remove PrefixUntil) and just add an optional parameter(the upstream parser) to PrefixUpTo. Doing this will make it work only in the FastPath and reduce the overall API size issue. But this will break the only slow path case we still have to figure out a better way to handle this case.

The implementation part can be improve too with a more clear and simpler approach. I will try you parser(because it’s seems to be way more cleaner than mine) and see if it could fit, pass the tests.

FWIW I think this should be added to PrefixThrough too.

Thanks,

mackoj:
Hi, @iampatbrown thanks for responding,

Maybe I'm missing something but in your example, I don't understand how the example below could work or event this test `testMoreComplexExempleFastPath`.

```swift
let separatorParser = Parse {
  "["
  Int.parser()
  Optionally {
    ";"
    Int.parser()
  }
  "m"
}

let lineParser = PrefixUntil("[") { separatorParser }
var input = "[08:45:33]Hello[31;0mworld[0m\n"[...]
try line.parse(&input)  // "[08:45:33]Hello"
input                             // "[31;0mworld[0m\n"
```

I did try a lot of approaches from yours(that is actually in production in my but does not cover all cases) to re-implement Many or do a recursive parser(https://github.com/pointfreeco/swift-parsing/discussions/152) even the Arithmetic(https://github.com/pointfreeco/swift-parsing/blob/main/Sources/swift-parsing-benchmark/Arithmetic.swift) road without success. But I found that the way it is done in this PR allows for expressing the parsing rule more clearly.

The goal of PrefixUntil is to consume an input until another parser(that cannot be represented as a simple string) succeed to extract the content of provided parser later.

I would love to know how to make this without adding a new parser to swift-parsing.

Thanks,

iampatbrown:
Hey @mackoj,

For the line parser, I believe something similar to @tgrapperon's example would work. 

```swift
Consumed {
  Many(1...) {
    PrefixUpTo("[")
  } separator: {
    Not { separatorParser }
    "["
  } terminator: {
    Peek { separatorParser }
  }
}
```

mackoj:
I will try your approach @iampatbrown thanks for taking the time to respond.

mackoj:
I'll close this for the moment and redo a PR with the work in PrefixUpTo

mackoj:
Hi,

I won't have the time to work on this for some time so if you need this operator here it is https://gist.github.com/mackoj/22f5709e518bc660c2d685f7a86fd163.

PR #245, opened 2022-07-17
Bring the coalescing operator `??` to parsers

tgrapperon:
Hello!
This PR adds two overloads for the nil-coalescing operator to sugar `.replaceError(with:)`. 
It should improve reachability of `ReplaceError` parsers that are very frequently needed.
 
If the left/upstream parser fails, it falls back with the right/default value. If `P<Output>` is some upstream parser and `default` an `Output` value, the overall parser always returns `default` when the upstream `P` fails in some way:
```swift
P<Int> ?? 1   P fails, parses 1
P<Int?> ?? 1  P fails, parses 1
P<Int>? ?? 1  P fails or is nil, parses 1
```
I think this is what users would naturally expect.

It shouldn't impact existing parsers.

We could maybe expand/update the documentation around `ReplaceError` or `Optionally`. I think we can also document these overloads if they appear in docc. I'll update this PR with more documentation commits, should it go further.

### DIFF
--- a/Sources/Parsing/ParserPrinters/ReplaceError.swift
+++ b/Sources/Parsing/ParserPrinters/ReplaceError.swift
@@ -85,3 +85,19 @@ extension Parsers.ReplaceError: ParserPrinter where Upstream: ParserPrinter {
     }
   }
 }
+
+@inlinable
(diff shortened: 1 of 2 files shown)


tgrapperon:
I'll close for now.

PR #249, opened 2022-09-13
Remove Vending Xcode schemes

mackoj:
Hi,

Would you consider removing vending Xcode schemes? Like you did in [TCA](https://github.com/pointfreeco/swift-composable-architecture/discussions/1371)?

Thanks,


stephencelis:
@mackoj We're definitely down to do this! Sorry if getting CI was a pain. Will try to look into this soon.

PR #250, opened 2022-09-14
lower restriction on End parser from Collection to Sequence

JaapWijnen:
Lowers the restriction on the End parser from Collection to Sequence.


JaapWijnen:
@stephencelis Is there any way to see if this has any impact on performance? (I don't expect it to but would be nice to double check)

stephencelis:
@JaapWijnen I doubt it, but given that Brandon is traveling at the moment and this PR isn't urgent, I might wait to discuss with him. Thanks for taking the time!

stephencelis (approved):
Thanks!

PR #255, opened 2022-09-30
Remove package resolved

### LINKED ISSUE #251: Docs action is failing
Hey, just noticed the docs gh action is failing therefore there aren't any docs generated for 0.10.0

JaapWijnen:
Tested and this seems to fix #251 
Are you ok with removing Package.resolved?


stephencelis:
@JaapWijnen Thanks for looking into this! It was helpful in getting us to track down the issue. I think we want `Package.resolved` checked in for this repo (and others) since it creates a more stable CI environment.

So instead we updated the docs CI job to discard anything that changes in the working tree:

https://github.com/pointfreeco/swift-parsing/commit/c2c6d38d7a69e7ac6339aa8cd70c3bfb97f5e6df

Going to close this in favor of the above, but thanks again! Your PR helped us figure out the problem more quickly :smile:

JaapWijnen:
@stephencelis Great! Yeah that was the main reason I made it :) Great! thanks for fixing it :)

PR #256, opened 2022-10-22
Allow match after `Prefix` with max count

oskarek:
The printing for the `Prefix` `ParserPrinter` fails whenever it detects that the next element printed matches its predicate. However, this should be allowed in the special case that the parser _has a maximum count_, and _the value to print has exactly that count_. This is because that would have stopped the parsing before the next element, regardless of whether it matches the predicate or not.


stephencelis (approved):
Good catch! Thanks for taking the time to PR and test!

PR #277, opened 2023-01-13
Extend CaseIterableRawRepresentable to FixedWidthInteger

ytyubox:
#276

> Hi @ytyubox, good catch!
> 
> Rather than making the initializer public, what if we made the parser work with all `FixedWidthInteger` types? Then we could all the various integer types.
> 
> If you have the time you want to open a PR to do that? If not we can take care of it.


mbrandonw (approved):
Thanks!

PR #285, opened 2023-02-11
Library evolution

abel3cl:
When building with flag, 
`@inlinable init` do not compile with the error:
Screenshot 2023-02-11 at 17 14 59

However marking `@usableFromInline` a `public init` is not allowed. 

Screenshot 2023-02-11 at 17 15 31

This PR creates an internal `init` as suggested by the first error that gets called from the `public init`
 
Refs: 
 - https://github.com/pointfreeco/swift-parsing/discussions/284
 - https://github.com/pointfreeco/swift-composable-architecture/pull/1443

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,3 +50,16 @@ jobs:
       - uses: MaxDesiatov/swift-windows-action@v1
         with:
           swift-version: "5.5.1"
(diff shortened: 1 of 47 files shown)


danthorpe:
This is great @abel3cl , we've been looking to get this working too!

PR #287, opened 2023-02-21
Add Backtrack parser.

mbrandonw:
Due to the discussions in #286 we realized that there is no way to explicitly backtrack just a single parser since `OneOf` does not undo the consumption of the last parser run. We think we want to keep that behavior, but think also explicit backtracking can be good, so we are adding a dedicated `Backtrack` parser.

### DIFF
--- a/Sources/Parsing/Documentation.docc/Articles/Backtracking.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Backtracking.md
@@ -70,7 +70,7 @@ It backtracks the input to its original value when the parser fails so that late
 The only time you need to worry about explicitly backtracking input is when making your own
 ``Parser`` conformances. As a general rule of thumb, if your parser recovers from all failures
 in the `parse` method then it should backtrack the input to its state before the error was thrown.
-This is exactly how ``OneOf``, ``Backtrack``, ``Optionally``, ``Not`` and ``Parser/replaceError(with:)`` 
+This is exactly how ``OneOf``, ``Backtracking``, ``Optionally``, ``Not`` and ``Parser/replaceError(with:)`` 
 work.
 
 ## Performance
--- a/Sources/Parsing/Documentation.docc/Articles/Backtracking.md
+++ b/Sources/Parsing/Documentation.docc/Articles/Backtracking.md
@@ -70,7 +70,7 @@ It backtracks the input to its original value when the parser fails so that late
 The only time you need to worry about explicitly backtracking input is when making your own
 ``Parser`` conformances. As a general rule of thumb, if your parser recovers from all failures
 in the `parse` method then it should backtrack the input to its state before the error was thrown.
-This is exactly how ``OneOf``, ``Backtrack``, ``Optionally`` and ``Parser/replaceError(with:)`` 
+This is exactly how ``OneOf``, ``Backtrack``, ``Optionally``, ``Not`` and ``Parser/replaceError(with:)`` 
 work.
 
 ## Performance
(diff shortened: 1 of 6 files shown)


stephencelis (approved):
Looks good! Only thought is to bikeshed the name a bit:

```swift
Backtracking { … }
WithBacktracking { … }
// ???
```

mbrandonw:
Doesn't seem like we use `-ing` suffixes for other parsers, e.g. we don't do `Skipping`, `Peeking`, etc. Makes me think an imperative noun is secretly our style for these things.

Sources/Parsing/Documentation.docc/Articles/Backtracking.md
```diff
@@ -70,7 +70,8 @@ It backtracks the input to its original value when the parser fails so that late
 The only time you need to worry about explicitly backtracking input is when making your own
 ``Parser`` conformances. As a general rule of thumb, if your parser recovers from all failures
 in the `parse` method then it should backtrack the input to its state before the error was thrown.
-This is exactly how ``OneOf``, ``Optionally`` and ``Parser/replaceError(with:)`` work.
+This is exactly how ``OneOf``, ``Backtrack``, ``Optionally`` and ``Parser/replaceError(with:)``
```
tgrapperon, line 73:
```suggestion
This is exactly how ``OneOf``, ``Backtrack``, ``Optionally``, ``Not`` and ``Parser/replaceError(with:)`` 
```

tgrapperon:
This seems to be a superior solution than making `OneOf` behavior backtrack for any parser. It doesn't impact performance and behavior, and provides a dedicated tool to implement backtracking in a much declarative way.

iampatbrown:
@mbrandonw I'd be leaning towards `Backtracking { … }` or `WithBacktracking { … }` as well. I feel in the case of `Skip` and `Peek` it's describing the successful behaviour. This one seems a bit different
`Peek { parser }` - peek the parser's output
`Skip { parser }` - skip the parser's output
`Backtrack { parser }` - backtrack the parser's output ??? 

Some other names to consider are `Lazy`, `Optionally`, `ReplaceError` and `OverridePrinting`...
Not really sure what's best though and `Backtrack` is fine too :)

When I see `Backtracking` or `WithBacktracking` I read them as:
`Backtracking { parser }` - this parser is now a backtracking parser
`WithBacktracking { parser }` - this parser now has backtracking

tgrapperon:
As an alternative, it could also be presented as a modifier, which is what it semantically is:
```swift
Parse { … }.backtracking()
```

mbrandonw:
@iampatbrown Very good points! I'm back on board with `Backtracking`/`WithBacktracking`, but not sure which 😅

stephencelis:
I guess I'd lean `Backtracking` since it's shorter and doesn't evoke all the `with*` functions in Swift.

iampatbrown:
> I guess I'd lean Backtracking since it's shorter and doesn't evoke all the with* functions in Swift.

@stephencelis I'd be the same. I think it sort of echos what's in the docs too:

> the backtracking `OneOf` work

which reads as  "the backtracking parser work"

mbrandonw:
Great, I've just pushed a rename to `Backtracking`. Ubuntu 18.04 seems to be having trouble on CI, but let's see how this latest run fares.

Sources/Parsing/Documentation.docc/Articles/Backtracking.md
```diff
@@ -70,7 +70,8 @@ It backtracks the input to its original value when the parser fails so that late
 The only time you need to worry about explicitly backtracking input is when making your own
 ``Parser`` conformances. As a general rule of thumb, if your parser recovers from all failures
 in the `parse` method then it should backtrack the input to its state before the error was thrown.
-This is exactly how ``OneOf``, ``Optionally`` and ``Parser/replaceError(with:)`` work.
+This is exactly how ``OneOf``, ``Backtrack``, ``Optionally``, ``Not`` and ``Parser/replaceError(with:)``
```
iampatbrown, line 73:
```suggestion
This is exactly how ``OneOf``, ``Backtracking``, ``Optionally``, ``Not`` and ``Parser/replaceError(with:)`` 
```

PR #289, opened 2023-03-24
Swift 5.8 Support

stephencelis:
To prepare for the release of Xcode 14.3 and Swift 5.8, Parsing must make some breaking changes. This PR overhauls the `ParserBuilder` and `OneOfBuilder` to be compatible with Swift 5.8's [new result builder implementation](https://forums.swift.org/t/improved-result-builder-implementation-in-swift-5-8/63192). Apple's decided to make breaking changes to result builders to improve their performance, and so we must adapt accordingly. Because of this, upgrading a project using Parsing to Swift 5.8 may not be trivial, but we'll do our best to describe the what, why, and how in an [associated discussion](https://github.com/pointfreeco/swift-parsing/discussions/290).


stephencelis:
Read more about this PR here: https://github.com/pointfreeco/swift-parsing/discussions/290

mbrandonw (approved):

PR #291, opened 2023-03-26
Downstream ParseBuilder should use Downstream.Input as Input

KaiOelfke:
I'm currently trying to migrate some parsing code to the 5.8 branch. I think I found a bug, but it could also be a misunderstanding with the latest changes.

The ParseBuilder is for Downstream so it should use it’s input. Or Upstream.Output as pipe runs the downstream parser on the output of the upstream parser. Which is why Upstream.Output == Downstream.Input.

With the 5.8 changes pipe uses the wrong input here. This only causes issues, if Upstream and Downstream have a different Input type.

The test can probably be simplified.

### DIFF
--- a/Sources/Parsing/ParserPrinters/Pipe.swift
+++ b/Sources/Parsing/ParserPrinters/Pipe.swift
@@ -31,7 +31,7 @@ extension Parser {
   /// - Parameter downstream: A parser that parses the output of this parser.
   /// - Returns: A parser that pipes this parser's output into another parser.  @inlinable
   public func pipe<Downstream>(
-    @ParserBuilder<Downstream.Input> _ build: () -> Downstream
+    @ParserBuilder<Output> _ build: () -> Downstream
   ) -> Parsers.Pipe<Self, Downstream> {
     .init(upstream: self, downstream: build())
   }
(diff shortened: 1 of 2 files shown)


Sources/Parsing/ParserPrinters/Pipe.swift
```diff
@@ -31,15 +31,15 @@ extension Parser {
   /// - Parameter downstream: A parser that parses the output of this parser.
   /// - Returns: A parser that pipes this parser's output into another parser.  @inlinable
   public func pipe<Downstream>(
-    @ParserBuilder<Input> _ build: () -> Downstream
+    @ParserBuilder<Downstream.Input> _ build: () -> Downstream
```
tgrapperon, line 34: I feel that `Output` would be slightly easier on the compiler in this position. `Self` generics are fully resolved at this staged, whereas `Downstream` aren't.
I'm a little surprised that it's working with `Downstream.Input`, because we're back in the chicken and egg situation, but I guess that the context provides enough information to resolve types here (I think because it knows that it'll returns a `Pipe` and this locks `Downstream.Input` to `Upstream.Output`).

KaiOelfke: That makes sense. I changed it. Thank you.

stephencelis:
Thanks!

PR #305, opened 2023-07-24
Improve `Date` parser demo

stephencelis:
Incorporates a helper from @juri's [parser](https://github.com/juri/Parse3339) to improve things in our demo a bit.

Benchmarked against Parse3339 [here](https://github.com/pointfreeco/swift-parsing/commit/fd4acb45111c9d696832d96023179cbe4adbc2a4) and it's nice to see how well the relatively naive combinators performed by comparison:

```
name                      time         std        iterations
------------------------------------------------------------
Date.Parser                3583.000 ns ±  18.18 %     355076
Date.Parse3339             3083.000 ns ±  19.16 %     448056
Date.DateFormatter        21708.000 ns ±   7.24 %      63634
Date.ISO8601DateFormatter 26958.000 ns ±   5.17 %      51322
```

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,52 +1,41 @@
 {
-  "object": {
-    "pins": [
-      {
-        "package": "swift-argument-parser",
-        "repositoryURL": "https://github.com/apple/swift-argument-parser",
-        "state": {
-          "branch": null,
-          "revision": "6b2aa2748a7881eebb9f84fb10c01293e15b52ca",
-          "version": "0.5.0"
-        }
-      },
-      {
-        "package": "Benchmark",
-        "repositoryURL": "https://github.com/google/swift-benchmark",
-        "state": {
-          "branch": null,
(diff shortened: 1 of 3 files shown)


mbrandonw (approved):

BobbyRohweder:
I don't think you want to use the `tm(...)` implementation from Parse3339 because it is very slow.
In the Parse3339, the `parse(s).components.date` is the faster string to date helper.

I also left that question / issue in the repo: https://github.com/juri/Parse3339/issues/7
My assumption was that is was only in there to get a benchmark for "Parse with Parse3339 (Unix time)" and compare it against "Parse with Parse3339 (DateComponents)"

stephencelis:
@BobbyRohweder I'm not sure how you're measuring it, but `DateComponents.date` is quite a bit slower:

```
Date.Parser (tm)                   3583.000 ns ±  18.18 %     355076
Date.Parser (DateComponents.date) 10750.000 ns ±  11.59 %     120687
```

PR #314, opened 2023-09-10
Improve JSON parser demo

oskarek:
There are a couple issues in the JSON parser demo (recently discussed in pointfree [this pointfree Slack thread](https://pointfreecommunity.slack.com/archives/C04L57GCR99/p1694082559893489)):
- Parsing characters built up of more than one utf-8 code point, such as `再`, fails.
- The literal unicode code point format is faulty, lacking a literal `u` character before the numbers.

This PR fixes those issues, as well as extends the JSON parser to fully conform to the json standard. This means adding support for handling UTF-16 surrogate pairs, and exactly match the JSON specification's number format, instead of relying on `Double.parser()` which isn't exactly the same.

### DIFF
--- a/Sources/swift-parsing-benchmark/JSON.swift
+++ b/Sources/swift-parsing-benchmark/JSON.swift
@@ -4,16 +4,16 @@ import Parsing
 
 /// This benchmark shows how to create a naive JSON parser with combinators.
 ///
-/// It is mostly implemented according to the [spec](https://www.json.org/json-en.html) (we take a
-/// shortcut and use `Double.parser()`, which behaves accordingly).
+/// It is implemented according to the [spec](https://www.json.org/json-en.html).
 let jsonSuite = BenchmarkSuite(name: "JSON") { suite in
   #if swift(>=5.8)
     struct JSONValue: ParserPrinter {
       enum Output: Equatable {
         case array([Self])
         case boolean(Bool)
+        case float(Double)
+        case integer(Int)
         case null
-        case number(Double)
         case object([String: Self])
         case string(String)
       }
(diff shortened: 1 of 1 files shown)


Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -4,16 +4,16 @@ import Parsing
...
 let jsonSuite = BenchmarkSuite(name: "JSON") { suite in
   #if swift(>=5.8)
     struct JSONValue: ParserPrinter {
       enum Output: Equatable {
         case array([Self])
         case boolean(Bool)
+        case float(Double)
+        case integer(Int)
         case null
-        case number(Double)
```
stephencelis, line 16: Can you explain this change? I was under the impression that JSON/JavaScript only has the concept of a "number" and doesn't distinguish integer and float values.

stephencelis (approved):
Thanks for the improvements here! Just one small thing to address before merging, I think.

Sources/swift-parsing-benchmark/JSON.swift
oskarek: Hmm yes, true. JSON only has the concept of a _number_, which is a sequence of digits, optionally followed by a decimal point and a fractional part. Now, if we model this as just a `Double` in Swift, we lose information about the number. `1` and `1.0`, for example, becomes indistinguishable. By modeling numbers that _does_ contain a fractional part as a `Double`, and numbers that doesn't as an `Int`, we at least keep that information, which I think is often desirable in Swift.

To not lose _any_ information we could model the numbers as a `String` (`case number(String)`), but that seems less convenient in practice. I think just separating numbers with and without any fractional part strikes a good balance of what information to keep about the numbers. This is at least how I'm thinking about this :)

oskarek:
@stephencelis how do you feel about getting this merged?

Sources/swift-parsing-benchmark/JSON.swift
oskarek: Sure, that's another valid option 🙂 I do think it's convenient, though, to have actual Swift number types (Int and Double in the case of my suggestion) as return types, if you would ever use this JSON parsing in practise.

oskarek: But at this point anything that gets this PR merged sometime would be nice 😁

oskarek:
@stephencelis I have updated the PR now, and removed the `Int`/`Double` separation

stephencelis:
@oskarek Thanks! Just one more thing came to mind.

Sources/swift-parsing-benchmark/JSON.swift
```diff
@@ -24,14 +23,53 @@ let jsonSuite = BenchmarkSuite(name: "JSON") { suite in
           JSONObject().map(.case(Output.object))
           JSONArray().map(.case(Output.array))
           JSONString().map(.case(Output.string))
-          Double.parser().map(.case(Output.number))
+          JSONNumber().pipe {
+            Double.parser().map(.case(Output.number))
+          }
```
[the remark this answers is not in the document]
oskarek: @stephencelis what do you think about this: ce9b709cd ?

PR #334, opened 2023-12-19
Fix the Prefix parser documentation

pinkjuice66:
Removed documentation for an unused initializer parameter in the Prefix parser.


stephencelis (approved):
Thanks!

Issue #349, opened 2024-08-17
Example in Getting started don't work - ambiguous use of 'buildExpression'


oschrenk:
Totally a possibility that I am missing something as I am a beginner Swift programmer, but the example on https://pointfreeco.github.io/swift-parsing/main/documentation/parsing/gettingstarted/

won't compile

```
  let parser = Parse {
    Int.parser()
    ","
    Bool.parser()
  }
```

will result in:

```
Building for debugging...
~/Project/Parser.swift:6:5: error: ambiguous use of 'buildExpression'
    Int.parser()
    ^
/Users/oliver/Projects/personal/swift/plan.swift/.build/checkouts/swift-parsing/Sources/Parsing/Builders/ParserBuilder.swift:533:22: note: found this candidate
  public static func buildExpression<P: Parser>(_ expression: P)
                     ^
/Users/oliver/Projects/personal/swift/plan.swift/.build/checkouts/swift-parsing/Sources/Parsing/Builders/ParserBuilder.swift:544:22: note: found this candidate
  public static func buildExpression<P: Parser>(_ expression: P) -> P
                     ^
error: emit-module command failed with exit code 1 (use -v to see invocation)
~/Project/Parser.swift:6:5: error: ambiguous use of 'buildExpression'
    Int.parser()
    ^
/Users/oliver/Projects/personal/swift/plan.swift/.build/checkouts/swift-parsing/Sources/Parsing/Builders/ParserBuilder.swift:533:22: note: found this candidate
  public static func buildExpression<P: Parser>(_ expression: P)
                     ^
/Users/oliver/Projects/personal/swift/plan.swift/.build/checkouts/swift-parsing/Sources/Parsing/Builders/ParserBuilder.swift:544:22: note: found this candidate
  public static func buildExpression<P: Parser>(_ expression: P) -> P
                     ^
~/Project/Parser.swift:6:5: error: ambiguous use of 'buildExpression'
    Int.parser()
    ^
/Users/oliver/Projects/personal/swift/plan.swift/.build/checkouts/swift-parsing/Sources/Parsing/Builders/ParserBuilder.swift:533:22: note: found this candidate
  public static func buildExpression<P: Parser>(_ expression: P)
                     ^
/Users/oliver/Projects/personal/swift/plan.swift/.build/checkouts/swift-parsing/Sources/Parsing/Builders/ParserBuilder.swift:544:22: note: found this candidate
[...]

PR #350, opened 2024-09-16
Fixed Playground compilation error

mattsrobot:
Playground example had a compilation error due to some changes in Swift Parsing.

Using the ParsePrint with input initialiser fixes the problem

Related issue: https://github.com/pointfreeco/swift-parsing/issues/349


stephencelis (approved):
Thanks!

PR #353, opened 2024-10-09
Fix a typo

RoBo-Inc:


mbrandonw (approved):
Good catch @RoBo-Inc, thanks!

PR #354, opened 2024-10-09
Don't use input.count in PrefixThrough and PrefixUpTo for speedup.

BjornRuud:
I have made a simple HTML tokenizer and noticed using Instruments that `PrefixThrough` was spending a lot of time in `Substring.count`, which in turn was calling `distance(from:to:)`. I assume this is because the character count of a substring isn't known until calculated due to how Unicode works. This patch replaces the substring count check with a match end index check instead.

I was unable to make sense of the results from the benchmark app, but in my HTML project (link provided) there was a substantial speedup. In that repo there is a benchmark app in the branch `swift-parsing`, and you can switch between swift-parsing versions in `Package.swift`.

https://github.com/BjornRuud/HTMLLexer.git


stephencelis (approved):
Good find, thanks!

BjornRuud:
The CI Ubuntu test failure is the same error I get for both the benchmark and tests with Swift 6.

PR #356, opened 2024-10-11
Fix a mistake in docs

RoBo-Inc:
The most important change is '123 ' --> ' 123' (the whitespace is in front, not behind).
That was quite misleading 😃


stephencelis (approved):
Thanks!

PR #357, opened 2024-10-12
Pitch: string trimming feature

RoBo-Inc:
Using the `trimming` function results in more concise a and nice-looking code compared to achieving the same effect by manipulating the `Whitespace()` parsers. Besides, it's convenient and intuitive to use this kind of function in such a way and context.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,33 +10,33 @@ on:
   workflow_dispatch:
 
 jobs:
-  macos_tests:
-    runs-on: macos-12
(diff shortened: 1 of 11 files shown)


RoBo-Inc:
Probably it's better to extend the existing `Conversions.SubstringToString` with the functionality, rather then to create a new conversion which refer to "fake" `TrimmedString` type 🙄

RoBo-Inc:
Probably, I should have put this in the "Discussion" section instead 🙄

PR #367, opened 2025-01-10
Removes the restriction on the amount of non-void Parsers that can be used in a Parserbuilder

coenttb:
This pull request eliminates the need for all Take*n overloads except for Take2 while maintaining full backward compatibility.

The Take2.Map parameter pack construct is a deliberate tradeoff. Ideally, we would prefer to use a `Conversion` with a `MapConversion` for a more elegant solution. However, consider the following hypothetical `TupleConversion`:
```swift
public struct TupleConversion<each T1, T2>: Conversion {
    public typealias Input = ((repeat each T1), T2)
    public typealias Output = (repeat each T1, T2)

    public func apply(input: ((repeat each T1), T2)) -> (repeat each T1, T2) {
        let (first, second) = input
        return (repeat each first, second)
    }

    public func unapply(output: (repeat each T1, T2)) -> ((repeat each T1), T2) {
        guard let tuple = output as? ((repeat each T1), T2) else {
            fatalError("Could not convert output tuple type")
        }
        return tuple
    }
}
```

While functional, this approach depends on macOS 14 for its use of parameter packs in generics, which makes it incompatible with earlier versions. As a result, introducing Take2.Map strikes a practical balance: it ensures backward compatibility while adding only minimal complexity to the surface area.

For verification, refer to `VariadicTests.swift`, which includes passing tests.

PS: I recommend exploring further if and to what extent the existing Map infrastructure can be leveraged over Take2.Map.

### DIFF
--- a/Sources/Parsing/Builders/ParserBuilder.swift
+++ b/Sources/Parsing/Builders/ParserBuilder.swift
@@ -127,90 +127,28 @@ public enum ParserBuilder<Input> {
   where P0.Input == Input, P1.Input == Input {
     .init(accumulated, next)
   }
-
+  
   @inlinable
   public static func buildPartialBlock<P0, P1>(accumulated: P0, next: P1) -> SkipSecond<P0, P1>
   where P0.Input == Input, P1.Input == Input {
     .init(accumulated, next)
   }
-
+  
   @_disfavoredOverload
-  @inlinable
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Thanks for looking into this! I'll bump the package file on `main` after merging since parameter packs require 5.9, but that's been around for awhile now.

Feel free to experiment with your `TupleParser` using compiler conditionals or availability checking in another PR if you think it'll improve performance on platforms that support it!

PR #370, opened 2025-01-16
Depend on CasePathsCore

stephencelis:
This is a breaking change, which means it should only be released in a 1.0 major update.


stephencelis:
Now that SwiftSyntax prebuilts are a thing, we're moving away from this `Core` module pattern.

PR #382, opened 2025-09-04
WIP: Migrate to Swift 6 and Strict Concurrency

floormatgen:
## Motivation
Using the Swift 6 language mode in conjunction the Parsing package requires adding ``@preconcurrency`` to all import statements to suppress errors related to concurrency, especially global parsers using ``static let``.

## Changes
- Enabled [``StrictConcurrency``](https://www.swift.org/documentation/concurrency), [``ExistentialAny``](https://docs.swift.org/compiler/documentation/diagnostics/existential-any) and [``MemberImportVisibility``](https://docs.swift.org/compiler/documentation/diagnostics/member-import-visibility/) on all targets
- Added a new target ``Swift6Tests`` for testing changes due to migration.
- Added Sendable to concrete Parsers and Conversions that can be Sendable
- Added conditional Sendable to Parsers and Conversions that take other parsers
- Added ``@preconcurrency`` to Parsers that have **transform** closures (e.g. ``Map``)
- Added ``@preconcurrency`` to ``ParsingError.Context`` and ``PrintingError.Context``
- Adopted [``ExistentialAny``](https://docs.swift.org/compiler/documentation/diagnostics/existential-any) to   ``protocol`` types.

## Future Directions
- Parsers that have transform closures can and should be marked ``Sendable``, as long as the transform closure is ``Sendable``. Currently this can be done with ``@Sendable`` but this makes the closure **always** ``@Sendable``, which affects source compatibility. If there was a way to mark a type as ``Sendable`` when one of its closures are ``Sendable``, this would be ideal.
    - We could also just mark the transform closures always ``@Sendable``. Even though this is a reasonable assumption for most parsers, this may not apply to some clients.
        - If we mark those ``Parser`` ``init``s  with ``@preconcurrency``, the closure's Sendability will only be checked if the client has ``StrictConcurrency`` or Swift 6 enabled, and only a warning will be generated if it is violated.
    - The cleaner syntax where an ``init`` is provided as a transform (i.e. ``.map(String.init)``) doesn't work if the transform closure is marked as ``@Sendable``, since Swift complains that the``init`` is not ``@Sendable``. This can be worked around by wrapping the ``init`` in another closure (i.e. ``.map { String.init($0) }``)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /Packages
 /*.xcodeproj
 xcuserdata/
+
+.vscode/
--- a/Makefile
+++ b/Makefile
@@ -45,4 +45,42 @@ generate-variadics:
 		--generate-one-ofs \
 		> Sources/Parsing/Builders/Variadics.swift
 
+
+# MARK: - Swift Compiler build tests
+
+DOCKER_CMD = \
+	docker run \
+	--rm \
+	-v $(PWD):/host \
+	-w "/host"
+
+TEST_SCRIPT = bash -c "\
+	cd ..; \
+	cp -r host container; \
+	cd container; \
+	swift package clean; \
+	swift test --parallel; \
+	"
+
+test-5-9:
+	$(DOCKER_CMD) \
+	swift:5.9.2 \
+	$(TEST_SCRIPT)
+
+test-5-10:
+	$(DOCKER_CMD) \
+	swift:5.10.1 \
+	$(TEST_SCRIPT)
+
+test-6-0:
+	$(DOCKER_CMD) \
+	swift:6.0.3 \
+	$(TEST_SCRIPT)
+
+test-latest:
+	$(DOCKER_CMD) \
+	swift:latest \
+	$(TEST_SCRIPT)
+
+
 .PHONY: benchmarks format generate-variadics test
--- a/Package.swift
+++ b/Package.swift
@@ -7,7 +7,7 @@ let swift6Settings: [SwiftSetting] = [
   .enableUpcomingFeature("StrictConcurrency"),
   .enableUpcomingFeature("ExistentialAny"),
   .enableUpcomingFeature("MemberImportVisibility"),
-//  .enableUpcomingFeature("NonisolatedNonsendingByDefault"),
+//  .enableUpcomingFeature("NonisolatedNonsendingByDefault"), // should not matter, package doesn't use Concurrency yet
 ]
 
 let package = Package(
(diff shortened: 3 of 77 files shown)


Mika5652:
@mbrandonw Hi, any chance to review and merge this PR in the near future? In our project, we also have many concurrency warnings coming from the Parsing library. 🙏🏻 Thanks!

mbrandonw:
Hi @Mika5652, as far as I know there's no real reason to go marking everything under the sun `Sendable` in this library. Can you please share a concrete example where you have concurrency warnings?

Mika5652:
> Hi @Mika5652, as far as I know there's no real reason to go marking everything under the sun `Sendable` in this library. Can you please share a concrete example where you have concurrency warnings?

We define parsers like this inside the TCA reducers:
```
static let deeplinkParser = Parse {
  PathComponent("path")
}
```
And it gives us the following warning:
`Static property 'deeplinkParser' is not concurrency-safe because non-'Sendable' type 'Parse<DeeplinkRequest, PathComponent<String>>' may have shared mutable state; this is an error in the Swift 6 language mode`

I marked `PathComponent` and also `DeeplinkRequest` `Sendable`, but `Parse` is not. The `PathComponent` is our custom implementation of `Parser` with a custom `parse` function.

I could rewrite it into computed property, but then I have to define concrete types manually, which is not what we want, of course.

```
static var deeplinkParser: Parse<DeeplinkRequest, PathComponent<String>> {
  Parse {
    PathComponent("path")
  }
}
```
Maybe we are using it wrong. To be honest, this logic was there before I started in the project and I don't have much experience with the Parsing library, yet.

EDIT: Basically this is the only warning we encountered when using Parsing library.

mbrandonw:
Hi @Mika5652, ah I see. This is really no different from SwiftUI:

```swift
import SwiftUI
let v = ForEach([1, 2], id: \.self) { index in
  EmptyView()
}
// 🛑 Let 'v' is not concurrency-safe because non-'Sendable' type 'ForEach<[Int], Int, EmptyView>' may have shared mutable state
```

And the solution isn't to just make everything `Sendable`.

Rather than storing parsers as globals or statics, you should be building parsers by conforming to the `Parser` (or `ParserPrinter`) protocols:

```swift
struct DeepLinker: ParserPrinter {
  var body: some ParserPrinter<…>
    OneOf {
      // …
    }
  }
}
```

And then you can do `DeepLinker().parse(…)`. Or, even better, because you have defined your own type you have the ability to make it sendable, so now you can do:

```swift
static let deepLinker = DeepLinker()
```

Mika5652:
@mbrandonw I'll definitely try that. Thanks for fast reply!

PR #391, opened 2026-06-01
Add SPI documentation target

davdroman:
## Summary
- Restore SPI configuration for the Parsing documentation target.
- Update current README documentation links to Swift Package Index URLs.
- Move the 0.10.0 documentation link to its matching Swift Package Index URL.

## Notes
Older README release links still point to the existing GitHub Pages documentation because their Swift Package Index equivalents return 404.

## Verification
- Confirmed SPI URLs for main, Getting Started, String Abstractions, and 0.10.0 resolve.
- Confirmed no pointfreeco.github.io main links remain in the repo.
- Ran git diff --check.

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,0 +1,4 @@
+version: 1
+builder:
+  configs:
+  - documentation_targets: [Parsing]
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):
Thanks!

PR #392, opened 2026-06-08
Modernize package settings.

### LINKED ISSUE #388: Parsing target should explicitly link CasePathsCore (link-time AnyCasePath undefined symbols on 0.14.1)
## Summary
The `Parsing` target uses `CasePathsCore` symbols at the type-descriptor level (via `Sources/Parsing/Conversions/Enum.swift` and the `extension CasePath: Conversion` API), but its `Package.swift` only declares a product dependency on the umbrella `CasePaths` library. In some build configurations this leaves `CasePathsCore.o` unlinked from `Parsing.framework`, surfacing as undefined-symbol errors in downstream consumers.
## Environment
- **swift-parsing:** `0.14.1`
- **swift-case-paths:** `1.x` (resolved transitively)
- **Swift toolchain:** `5.9` / `6.0`
- **Xcode:** `16.x`
- **Affected build:** large multi-package iOS app test plan

## Symptom
Undefined symbols for architecture arm64:
  "CasePathsCore.AnyCasePath<…>.… type descriptor", referenced from:
      … in Parsing.framework
ld: symbol(s) not found for architecture arm64
Reproducible on a large test plan; intermittent across DerivedData state and parallel-build settings, deterministic once the consumer reaches a certain dependency-graph size.

## Root cause
`Package.swift` declares:
```swift
.target(
  name: "Parsing",
  dependencies: [.product(name: "CasePaths", package: "swift-case-paths")]
)
```

CasePaths re-exports CasePathsCore, which is enough to compile, but SwiftPM does not always pull CasePathsCore.o into the linked Parsing.framework. When a downstream test bundle re-links Parsing against a slightly different transitive view of swift-case-paths, the CasePathsCore type descriptors are missing and the link fails.

## Proposed fix
Additive, manifest-only, patch-release-safe.
Add `CasePathsCore` as an explicit product dependency of the Parsing target, alongside CasePaths:
```swift
 .target(
   name: "Parsing",
-  dependencies: [.product(name: "CasePaths", package: "swift-case-paths")]
+  dependencies: [
+    .product(name: "CasePaths", package: "swift-case-paths"),
+    .product(name: "CasePathsCore", package: "swift-case-paths"),
+  ]
 ),
```

Properties of this fix
- Additive. CasePaths continues to be linked; consumers of the umbrella product are unaffected.
- No source changes. Sources/Parsing/** is untouched. The extension CasePath: Conversion API surface is preserved.
- No floor bump. swift-case-paths requirement stays from: "1.0.0".
- No test changes. Existing tests pass.
- Distinct from #370, 370 replaces CasePaths with CasePathsCore and migrates the public API to CaseKeyPath / @CasePathable — a breaking change correctly gated on 1.0. The fix proposed here is orthogonal: it only ensures the object file Parsing already needs is reliably linked.

## Evidence
The exact manifest patch above was applied to a vendored copy of swift-parsing 0.14.1 and the link-time errors disappear deterministically — https://github.com/lightspeed-hospitality/ios-pos/pull/6659.
> Note: Vendoring is not a viable long-term solution. Every consumer of a transitively-pinned swift-parsing collides on package identity (Conflicting identity for swift-parsing, which SwiftPM has flagged as escalating to an error in future versions). Picking up a patched upstream release would let consumers retire their forks.

## Ask
1. Confirm the diagnosis (or correct it).
2. If accepted, would you consider tagging the fix as 0.14.2 so consumers stuck on 0.14.x can pick it up without waiting for the 1.0 migration? Happy to open the PR if that helps.
---
Thanks for the great library!

mbrandonw:
Should fix #388

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.9
+// swift-tools-version:6.1
 
 import PackageDescription
 
@@ -41,3 +41,21 @@ let package = Package(
     ),
   ]
 )
+
+for target in package.targets {
+  target.swiftSettings = target.swiftSettings ?? []
+  target.swiftSettings?.append(contentsOf: [
+    .swiftLanguageMode(.v5),
+    .enableUpcomingFeature("ExistentialAny"),
+    .enableUpcomingFeature("ImmutableWeakCaptures"),
+    .enableUpcomingFeature("InferIsolatedConformances"),
+    .enableUpcomingFeature("InternalImportsByDefault"),
+    .enableUpcomingFeature("MemberImportVisibility"),
+    .enableUpcomingFeature("NonisolatedNonsendingByDefault"),
+  ])
+  #if compiler(>=6.4)
+    target.swiftSettings?.append(contentsOf: [
+      .treatAllWarnings(as: .error)
+    ])
+  #endif
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,35 +6,16 @@ on:
       - main
   pull_request:
     branches:
-      - '*'
+      - "*"
   workflow_dispatch:
 
 jobs:
-  macos15:
-    name: macOS (Swift 6)
-    runs-on: macos-15
+  macos:
+    name: macOS
     strategy:
       matrix:
-        xcode:
-          - '16.2'
-    steps:
-      - uses: actions/checkout@v4
-      - name: Select Xcode ${{ matrix.xcode }}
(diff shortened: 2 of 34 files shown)


Package.swift
```diff
@@ -41,3 +41,21 @@ let package = Package(
     ),
   ]
 )
+
+for target in package.targets {
+  target.swiftSettings = target.swiftSettings ?? []
+  target.swiftSettings?.append(contentsOf: [
+    .swiftLanguageMode(.v5),
```
mbrandonw, line 48: Gotta stay in Swift 5 mode for the time being...

stephencelis (approved):

